PR #6, opened 2019-02-04
Fix Electron error for Mac 

kyjus25:


adlerluiz:
@kyjus25 Thank you for your PR

PR #13, opened 2019-05-11
Automate settings for tray control and bug fixes.

mingjun97:
* Automate detect dark mode, elimate out corresponding options in settings.html
* AUtomate hide tray control options for windows/linux
* Fix bugs while swtich tray control(Mac) back to normal tray mode
* Add menu for Mac in order to enable sorts of Cmd+C/V/A
* Fix logic about the process to handle Cmd+W and Cmd+Q in MacOS

---------

**Known issue**:
An alert saying exited unexpectedly after hit Cmd+Q. This is a false positive alert. :)


mingjun97:
The alert might be my mac's misconfiguration. I see the Discord also raise this kind of alert.

PR #31, opened 2019-07-17
Improve window

pinkiesky:
* Titlebar settings
* [WIP] Create documentation for store
* Fix window size


rickinbsb:
Did you move the settings button somewhere when using system bar?

adlerluiz:
The settings button already have on menu right click on tray icon, but need some fixes

When !== nice

- Compensate height to control bar from webpage;
- Show notification when have a new version of YTMDesktop;

PR #37, opened 2019-07-30
* onFocus was missing.

austonpramodh:
on minimizing the window and opening it back didn't use to focus on the page, therefore it didn't use to take any inputs from the keyboard.


adlerluiz:
Thank you for PR

PR #51, opened 2019-10-09
Update README.md

garfunkel:
Add Arch Linux package information


garfunkel:
One thing is that I'm not super experienced with Electron - so I sorta guessed the dependencies for the AUR package.

PR #56, opened 2019-10-18
Implemented read youtube links from clipboard

ZeusAFK:
App reads clipboard every 1000 ms and plays youtube links.
Added package electron-clipboard-watcher.


adlerluiz:
@ZeusAFK Thank you for your contribution, please make the changes on your development branch of ytmdesktop, because the merge with master will result on error

Issue #67, opened 2019-11-27
Couldn't sign in


diegocade1:
Hi, is there a work around about this 'You are trying to sign in from a browser or app that doesn't allow us to keep your account secure.' message?, finally find the app I was looking for but I got stuck here
![image](https://user-images.githubusercontent.com/46531987/69763705-d364b280-114c-11ea-96c1-f4f5c45694bf.png)

mingjun97:
Which system are you using?

diegocade1:
> 
> 
> Which system are you using?

windows 10 64bit

diegocade1:
I have 2FA in my account, is that a problem?

mingjun97:
I have 2FA in my account too. It seems I don't have such problem. Do you have chrome installed in your system?

diegocade1:
> I have 2FA in my account too. It seems I don't have such problem. Do you have chrome installed in your system?

Yeah

adlerluiz:
@diegocade1 Have you the latest release?

https://github.com/ytmdesktop/ytmdesktop/releases

diegocade1:
I checked the version and its the latest, sadly I still have the issue

mingjun97:
Sorry, we can't reproduce your issue in our environment now. So that we can't fix it at this moment.
Maybe you can try the following suggestions:
1. Clean up cookies of YTMDesktop.
2. Check your internet connections to see whether you have proxy or DNS misconfiguration.
3. Try to login with another account to see if it occurs again.

Again, thanks for your feedback and love our application!

AdityaTD:
I could replicate this issue on Windows 10 64bit as well. https://i.imgur.com/wXuR5VI.png
Using v1.5.3 of this app, I also have 2FA on.

Just an idea, what if you used an external popup to log in so you could log in via Chrome or any other browser and it'd call back to the app with the login tokens which could be injected in the electron browser thus not having to face this issue at all?

diegocade1:
> 
> 
> Sorry, we can't reproduce your issue in our environment now. So that we can't fix it at this moment.
> Maybe you can try the following suggestions:
> 
>     1. Clean up cookies of YTMDesktop.
> 
>     2. Check your internet connections to see whether you have proxy or DNS misconfiguration.
> 
>     3. Try to login with another account to see if it occurs again.
> 
> 
> Again, thanks for your feedback and love our application!
I tried everything, still no luck

AdityaTD:
[...]

PR #72, opened 2019-12-05
Fix: Unsafe browser on Login with User Agent

AdityaTD:
A temporary fix to the issue #67.
This is a known issue on all Chromium-based browsers due to the recent changes that Google is rolling out. More info at https://security.googleblog.com/2019/04/better-protection-against-man-in-middle.html

The bug was replicable by almost all users on other Electron applications such as Google Play Music Desktop Application, etc.

Current Fix:
- Using the latest Mozilla Firefox's User Agent on the main browser window instead of the Chromium's default one.

Files Changed:
- main.js

### DIFF
--- a/main.js
+++ b/main.js
@@ -138,6 +138,9 @@ function createWindow() {
     }
   }
   mainWindow = new BrowserWindow(broswerWindowConfig);
+  mainWindow.webContents.session.setUserAgent(
+    "Mozilla/5.0 (Windows NT 6.1; WOW64; rv:54.0) Gecko/20100101 Firefox/71.0"
+  );
   const view = new BrowserView({
     webPreferences: {
       nodeIntegration: true,


rickinbsb:
Amazing, thanks Aditya

rickinbsb:
Can you update the commit message to use gitmoji? I believe :bug: or :lock: would fit

AdityaTD:
There you go :)

mingjun97:
Great works! Thanks!

aliocantinea:
I sorry I don't understand how to apply this fix that you have created... can you walk me through it? @AdityaTD

rickinbsb:
The fix is in the app itself, you just have to update to the new version when it comes out, today probably

diegocade1:
> 
> 
> The fix is in the app itself, you just have to update to the new version when it comes out, today probably

I just update the app but it still says is not secure

diegocade1:
I just checked if this portion of code was in the main branch

 'mainWindow.webContents.session.setUserAgent(
    "Mozilla/5.0 (Windows NT 6.1; WOW64; rv:54.0) Gecko/20100101 Firefox/71.0"
  );'

 beneath

 'mainWindow = new BrowserWindow(broswerWindowConfig);'

but it isn't....

AdityaTD:
It's in the development branch at the moment, it'll be merged into master soon I believe.

The fix basically makes google believe that we're logging in from a Firefox browser instead of a Chromium based browser which Google doesn't let login due to MITM attacks.

Once the update is out, you should be able to login properly, I've tested it multiple times.

PR #119, opened 2020-03-19
🌐 Korean translate update

ddarkr:
Added translation based on english. 👍
There is some paraphrase...

I love this application. thank you!


adlerluiz:
Thank you for using YTMDesktop and for your PR =D

Issue #137, opened 2020-04-23
starting reduced in the tray bar


chalard87:
Hi, thanks for this great app.
Could it be possible to  have a "start reduce in tray bar at launch" feature ? 
It start very well with windows 10 but always in a window, not reduced.
thanks,
JM

NNowakowski:
Implemented as part of #148

PR #148, opened 2020-05-02
:sparkles: Minimize on startup

NNowakowski:
Added settings to start the application mizimized

As per #137


adlerluiz:
Great, thanks for the PR

PR #174, opened 2020-06-01
💬 New strings

TotallyNotInUse:
Added new strings.

Also, there seems to be a red : at line 62, don't know why though.


rickinbsb:
the red `:` was because there was a missing `,` on line 61, @adlerluiz already fixed it tho

Issue #195, opened 2020-06-11
MacOS menu bar icon incorrect render/scaling


manastalukdar:
The menu bar icon is showing up incorrectly as shown in the screenshot below.

<img width="653" alt="Screenshot at Jun 11 12-21-17" src="https://user-images.githubusercontent.com/2075466/84430213-2fb98c00-abde-11ea-980c-5c39253c038d.png">

adlerluiz:
thanks, let's fix this, unfortunately we don't have mac to test

manastalukdar:
@adlerluiz , I am happy to test it out on a mac once a commit has been made.

adlerluiz:
@manastalukdar PR was done on development and can be tested

manastalukdar:
@adlerluiz , Confirming that it is now fixed with latest on `development`.

rickinbsb:
cool! this PR will come out on 1.11

adlerluiz:
Thank you for your feedback

PR #199, opened 2020-06-13
:bug: fix incorrect size on mac tray icon

zagoruev:
#195 
Apply scaling for tray icon on macOS
Also, slightly refactor code:
- Remove redunant set of icon in setShinyTray method as we already did it in createTray
- Move nativeImageIcon building to separate method to reuse it


markpermann:
looks like this bug (or a similar one?) is still a problem on 1.12.0:

image

PR #204, opened 2020-06-18
:sparkles: use monochrome, theme depending, tray-icon for macOS

zagoruev:
Changes sample


zagoruev:
**Also:**
:sparkles: Assing "Cmd+," shortcut to open settings on macOS

PR #212, opened 2020-06-24
:lipstick: Make next song notifications silent

Wani4ka:
Notifications about new song playing may distract. In GPM desktop, for example, such notifications are silent.


rickinbsb:
this is pretty good, thanks!

Issue #342, opened 2020-08-27
Enable contextIsolation for all renderers


MarshallOfSound:
- [x] I understand that **YTMDesktop have NO affiliation with Google or YouTube**.
- [x] I verified that there is no open issue for the same subject.

**Describe the bug**
As per the [Electron Security Guidelines](https://www.electronjs.org/docs/tutorial/security) `contextIsolation` should be enabled for all renderers.  This means every `BrowserWindow` and `BrowserView`.

Refs: https://github.com/ytmdesktop/ytmdesktop/blob/master/main.js#L223

The technically correct way of doing this is by exposing a minimal API surface area through the `contextBridge` module from Electron --> https://electronjs.org/docs/api/context-bridge

**To Reproduce**
N/A

**Expected behavior**
N/A

**Screenshots**
N/A

**Environment (please complete the following information):**
 * YTMDesktop version: ***v1.12.1***
 * OS: ***macOS***
 * OS version: ***10.15***
 * Arch: ***x64***
* Installation way: Any

Alipoodle:
Closing this PR in the move to YTMD v2

Issue #341, opened 2020-08-27
Disable `nodeIntegration` for remote content


MarshallOfSound:
- [x] I understand that **YTMDesktop have NO affiliation with Google or YouTube**.
- [x] I verified that there is no open issue for the same subject.

**Describe the bug**
As per the [Electron Security Guidelines](https://www.electronjs.org/docs/tutorial/security) `nodeIntegration` should be disabled for remote content.

Refs: https://github.com/ytmdesktop/ytmdesktop/blob/master/main.js#L223

Ideally `nodeIntegration` would be disabled on **all** windows for enhanced security

**To Reproduce**
N/A

**Expected behavior**
N/A

**Screenshots**
N/A

**Environment (please complete the following information):**
 * YTMDesktop version: ***v1.12.1***
 * OS: ***macOS***
 * OS version: ***10.15***
 * Arch: ***x64***
* Installation way: Any

Alipoodle:
Closing this PR in the move to YTMD v2

PR #344, opened 2020-08-27
🔒 enable security settings for the main window

MarshallOfSound:
These include `contextIsolation: true`, `sandbox: true`, `nodeIntegration: false` and `enableRemoteModule: false`

Refs: #343 #342 #341

Was actually surprisingly straight forward, most of the code worked out of the box. Not sure if I missed anything though so would be great if someone more familiar with the codebase gave it a whirl.

### DIFF
--- a/main.js
+++ b/main.js
@@ -239,9 +239,12 @@ function createWindow() {
 
     view = new BrowserView({
         webPreferences: {
-            nodeIntegration: true,
-            webviewTag: true,
-            enableRemoteModule: true,
+            nodeIntegration: false,
+            webviewTag: false,
+            enableRemoteModule: false,
+            contextIsolation: true,
+            sandbox: true,
+            nativeWindowOpen: true,
             preload: path.join(
                 app.getAppPath(),
                 '/src/utils/injectControls.js'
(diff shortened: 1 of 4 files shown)


main.js
```diff
@@ -239,9 +239,12 @@ function createWindow() {
 
     view = new BrowserView({
         webPreferences: {
-            nodeIntegration: true,
-            webviewTag: true,
-            enableRemoteModule: true,
+            nodeIntegration: false,
```
adlerluiz, line 242: When performing the test with these modifications, the buttons injected into the interface no longer appear.

image

MarshallOfSound: Hm they appeared on my local, let me double check.  We're there any errors in your devtools for the BrowserView?

adlerluiz: image

MarshallOfSound:
@adlerluiz Can you confirm you're on the latest version of my branch.  I just double checked and I definitely get all the injected UI 🤔  That errors look like you're not using my version of `injectControls.js` because I completely removed that `require` call 🤔

adlerluiz:
I cloned in another directory and now works correctly 😊

src/providers/settingsProvider.js
```diff
@@ -30,8 +31,45 @@ function setInitialValue(settingName, initialValue) {
 }
 
 function onDidChange(key, callback) {
-    store.onDidChange(key, (newValue, oldValue) => {
-        callback({ newValue: newValue, oldValue: oldValue })
+    return store.onDidChange(key, (newValue, oldValue) => {
+        callback({ newValue, oldValue })
```
adlerluiz, line 35: image

changing to callback(newValue, oldValue)

image

MarshallOfSound: Fixed 😄

PR #346, opened 2020-08-27
🔨 add github actions CI

MarshallOfSound:
Adds some basic CI to build all three platforms and store the artifacts for easy testing / validating of PRs.  This can be extended with linting / tests in the future.


MarshallOfSound:
This has the strange side-effect of moving from `npm` to `yarn` as npms `package-lock` does not work very well across platforms with optional dependencies.  `yarn` handles this significantly better (i.e. it actually works)

MarshallOfSound:
Also worth noting GH actions is completely free for public repositories so this is zero-cost 😄  🎉

adlerluiz:
> Also worth noting GH actions is completely free for public repositories so this is zero-cost 😄 🎉

Nice!

PR #374, opened 2020-09-11
:sparkles: improve volume controls

Nerogar:
I have improved the volume controls:
 - The 5% volume steps have been removed
 - When using the global volume shortcuts or by scrolling in the player bar or the mini player the volume is increased/decreased by 1.5dB instead of 5%
 - A new setting was added that allows the user to use their volume media keys to control only the player volume instead of the whole system volume


adlerluiz:
I liked its implementation, I'm just thinking about maybe in the playback settings add a button

```Switch volume control from '%' to 'dB'```

and the volume control shortcuts in terms of usability for the user

for example when activating this option
image

it will not be necessary this
image

I think of two options

1. Leave as is:
VolumeUp and VolumeDown always control the volume of the OS
Custom keys to control the volume of the app

2. Remove the volume control custom keys from the app, leave VolumeUp and VolumeDown for when the `` `Enable media volume keys``` option is:
on - Control OS volume
off - Control app volume

rickinbsb:
I'm not comfortable with hijacking volume keys, so I would go with the first option too

Nerogar:
Thanks for the feedback.

I will add a setting for the dB/% switch. I just wasn't sure if that is needed because the dB scale is clearly superior 😄.

For the second one: Most people probably don't use the OS volume and always leave it at a fixed value. The volume media keys don't get any use in this case. I added this because I couldn't get the media volume keys to work as a shortcut. They do show up in the ui, but they don't work:
image
Adding them like this:
image
Also didn't really work because I couldn't find a good place for the toggle to enable/disable these keys.

I would prefer the following solution, but I'm not sure if that is possible:
The volume Up/Down keys can be added as a normal shortcut. When they are set in the app they don't control the OS volume anymore.
Is that a solution you would accept?

adlerluiz:
no problem, I just want to have the discussion here so we can reach an agreement and you don't have any additional work to be undone later (I know how annoying this is =P)

1. I agree, the scale in dB is better, I believe that it is not necessary to make an option for this and leave it as the default.

2. Most people leave the volume fixed and will not have a problem with this modification, but precisely those who always use it, will complain here at some point, that's why every modification I like to let the person choose as he prefers.

I think we just need to change this name to make it more suggestive
image

something like 'media key controls only app volume'

Nerogar:
The dB/% toggle is something I will add. Letting the user choose is probably better.

For the volume button setting: I think I misunderstood your feedback. I thought you wanted me to remove that setting because hijacking these keys might confuse the user. I agree that not many people will use this feature, that's why i set the default to disabled. If renaming is all you want then that would be the easiest solution.

adlerluiz:
sorry I didn't explain it correctly, I was thinking and typing ...

summing up custom volume keys would be just to control the app, and volumeUp and down of the keyboard is by default to control the volume of the OS and if you activate the option, it will control the volume of the player.

because in the tests it even happened to me, I enabled the hijacking of the key, but I had the player open and paused, I went to listen to an audio on the Instagram web and I pressed the volume button on the keyboard and it didn't change anything .. then I remembered ... so this will be able to happen repeatedly with users

PR #389, opened 2020-09-16
✨ Add background color changes to match functionality of YouTube Music Android app.

TotalChris:
Additional Changes:

📱 Add style rules to injection script to define when the colors affect the UI elements and when they won't (Like if player is collapsed, but navigation bar is stickied).

---

The YouTube Music Android app has undergone a UI change this year which brought more Material-like Design elements to the player and other areas. This PR adds a function to ytmdesktop which dynamically changes the background colors of the player page when a song is playing. These match the main color of the album art for the song, and for the most part, line up with the colors that the Android app uses.

Color example 1

The measurements account for lighter colors by setting a fixed brightness and saturation. This album's pervasive color would normally be very bright, clashing with the UI text. Here, it has a fixed saturation of 70% and a fixed brightness of 25%.

Color example 2

The final example below is where the color engine struggles, as usually this album color is gold and not the shade of red that the chain's shadow casts.

Bad color example


TotalChris:
Setting to toggle this was just added.  It controls a Boolean attribute on the body element of the page. I modified the style rules to take effect contingent on that attribute being present. Since I used the InjectControls.js script to add the style rules, I find it easier to create a disabling mechanism rather than attempt to remove the rules completely.

adlerluiz:
@TotalChris Thank you for PR, great work

adlerluiz:
as you needed to add this in main.js i'm thinking of exposing the color also in the api and the miniplayer recovering its color, so we won't need to call the same functionality twice

adhillon192:
I would love to contribute to this!

Alipoodle:
@adhillon192 This has been merged into the codebase already!
It's not in the stable build though.

PR #405, opened 2020-09-23
✨ Catalan Translations

ArnyminerZ:
Added the translations for Catalan in `/locales`, not sure if they should be added anywhere else, if so, comment it in, and I'll be pleased to commit the necessary changes.


ArnyminerZ:
Moving request to development branch

PR #416, opened 2020-09-26
:art: Doing code maintenance

ArnyminerZ:
Increased `esversion` to `8` for having async functions.
Fixed unexpected type coercion equality checks.
Moved all inline function declarations from `function ()` to `() =>`.
Added semicolons to all the lines.
Added some TODO regarding unused functions and strange lines.
Reduced curly brackets presence.


TotalChris:
Big ups on this ✊

rickinbsb:
after adding this we need to make sure husky will enforce all of the changes on future commits

PR #450, opened 2020-10-13
This adds play and pause event handling to the mpris providers

### LINKED ISSUE #440: mpris/dbus interface doesn't allow to just pause or stop track
- [x] I understand that **YTMDesktop have NO affiliation with Google or YouTube**.
- [x] I verified that there is no open issue for the same subject.

**Is your feature request related to a problem? Please describe.**
I am using a script to pause all music players when I am in a video call. This script currently sends the "Pause" event to the player. On YTM this is mapped to "PlayPause", which means music starts playing if it wasn't playing before. 

**Describe the solution you'd like**
Support an "Pause" or "Stop" event and make it  separate from "PlayPause". You probably also want to do this for "Play"

**Describe alternatives you've considered**
An alternative for my script to first check the status and then conditionally do "PlayPause", but this is not necessary for any other player like GTM or Rhytmbox. 

**Additional context**
This is probably the context in the code: https://github.com/ytmdesktop/ytmdesktop/blob/master/src/providers/mprisProvider.js#L57-L59

This is how Rhythmbox maps them: https://github.com/GNOME/rhythmbox/blob/master/remote/dbus/rb-client.c#L107-L110

And GTM: https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/src/main/features/linux/mprisService.js#L33-L47

christofdamian:
Adding the play and pause events in addition to the play-pause
event. This makes it easier to pause the player externally if you
don't have access to the current status.

I am using an instance variable, because this.player.playbackStatus
doesn't seem to be available all of the time.

This fixes #440

JavaScript is not my first language, so I would be happy about any feedback.

### DIFF
--- a/main.js
+++ b/main.js
@@ -517,6 +517,9 @@ function createWindow() {
                 lastIsPaused = playerInfo.isPaused
 
                 discordRPC.setActivity(getAll())
+                if (isLinux()) {
+                    mprisProvider.setActivity(getAll())
+                }
 
                 if (!isMac() && !settingsProvider.get('settings-shiny-tray')) {
                     tray.updateTrayIcon(
--- a/src/providers/mprisProvider.js
+++ b/src/providers/mprisProvider.js
@@ -6,6 +6,7 @@ class Mpris {
         this._isInitialized = false
         this.player = undefined
         this._realPlayer = undefined //we'll need the infoPlayer later to be better able to track the time.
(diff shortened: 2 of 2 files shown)


christofdamian:
Sorry, closing this as it doesn't quite work yet. The current play status is not ending up in the callbacks.

PR #456, opened 2020-10-19
Allows raising of player

andia89:
This allows to open the window when the player is in the  notification area and you click on it to show it again (GNOME 3 for example).


adlerluiz:
hello, could you please switch to branch development

PR #457, opened 2020-10-19
Adds shutdown button

### LINKED ISSUE #418: Close button when tray icon is not available
- [x] I understand that **YTMDesktop have NO affiliation with Google or YouTube**.
- [x] I verified that there is no open issue for the same subject.

There are several Linux distros now that don't have tray icons anymore (mainly Gnome distros with the exception of Ubuntu), so I the tray icon will not show either if it is enabled. As far as I can see this means that there is no way (except the command line) to close the app completely if the keep running in background option is enabled. 

Would it be possible to have a "OFF" switch somewhere that really closes the app?

andia89:
This adds a button next to the settings button that allows to close the player completely. This is the best placement I found even though it might be better to add it to the popup menu that opens when you click your usericon. Unfortunately I was not able to do that, so if you want that there you will have to point me in the right direction

Fixes #418

### DIFF
--- a/main.js
+++ b/main.js
@@ -988,6 +988,13 @@ async function createWindow() {
         tray.setShinyTray()
     })
 
+    ipcMain.on('closed', (_) => {
+        mainWindow = null
+        if (process.platform !== 'darwin') {
+            app.quit()
+        }
+    })
+
     ipcMain.on('btn-update-clicked', () => {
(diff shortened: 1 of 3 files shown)


adlerluiz:
hello, could you please switch to branch development

andia89:
done on both of my PRs

PR #473, opened 2020-10-27
Update pt.json

smarquespt:


rickinbsb:
you forgot a comma on line 3

rickinbsb:
also, you are converting the brazilian portuguese translation to european portuguese, it would be better to have them as separate translations

smarquespt:
> you forgot a comma on line 3

Fixed

PR #478, opened 2020-10-29
Update pt.json, create pt-BT (previously pt.json)

smarquespt:


adlerluiz:
I see you created 2 PR, which one is the correct one, this one or #477 ?

smarquespt:
> I see you created 2 PR, which one is the correct one, this one or #477 ?

Both are, one is for development branch and another is for master

PR #483, opened 2020-10-31
Updated track-play and pause

XellaKat:
Added a check for the individualized play and pause commands so they work as expected and don't toggle the media state when not needed.

### DIFF
--- a/src/providers/companionServer.js
+++ b/src/providers/companionServer.js
@@ -378,7 +378,7 @@ function execCmd(cmd, value) {
             break
 
         case 'track-play':
-            if (infoPlayerProvider.getPlayerInfo().isPaused == true) {
+            if (infoPlayerProvider.getPlayerInfo().isPaused) {
                 ipcMain.emit('media-command', {
                     command: 'media-play-pause',
                     value: true,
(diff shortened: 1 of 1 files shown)


src/providers/companionServer.js
```diff
@@ -378,17 +378,21 @@ function execCmd(cmd, value) {
             break
 
         case 'track-play':
-            ipcMain.emit('media-command', {
-                command: 'media-play-pause',
-                value: true,
-            })
+            if (infoPlayerProvider.getPlayerInfo().isPaused == true) {
```
SaveliyKolesnikov, line 381: May be simplified to just  if (infoPlayerProvider.getPlayerInfo().isPaused)

src/providers/companionServer.js
```diff
@@ -378,17 +378,21 @@ function execCmd(cmd, value) {
...
+                })
+            }
             break
 
         case 'track-pause':
-            ipcMain.emit('media-command', {
-                command: 'media-play-pause',
-                value: true,
-            })
+            if (infoPlayerProvider.getPlayerInfo().isPaused == false) {
```
SaveliyKolesnikov, line 390: May be simplified to just  if (!infoPlayerProvider.getPlayerInfo().isPaused)

PR #491, opened 2020-11-08
Possibility of adding to a flatpak repository

linusrg1:
Its great that you offer an appimage for use, but in addition could you add a flatpak version as I know me and many other people dont like using snaps as they are not open source and they are maintained by Canonical who will probably never share the source code or allow anyone to create their own snap repos. If you could add a flatpak it would be really awesome!


adlerluiz:
we already have an issue with this request, we are studying how to create and create the flatpak

https://github.com/ytmdesktop/ytmdesktop/issues/306

PR #573, opened 2021-01-10
Use mpris only as soon as media is playing

andia89:
This is a fix that makes the mpris service startup only when necessary (a fix for LInux). 

Before the fix the controls showed upas soon as the program started which makes no sense since there is no media playing yet  (I only tried GNOME but I strongly suspect also other DEs). Now the player shows up in the notification area only after music starts playing


rickinbsb:
Thank you for this!

PR #602, opened 2021-01-30
Fix error not defined message at start up

nandafirdaus:
Hi there,

I got exception "error is not defined" when the app open.
image

Made this simple changes to fix the issues. Also adding condition in the so the code doesn't throw exception when tray is null.


mingjun97:
Thanks for contribution!

PR #606, opened 2021-02-02
fix: Adding missing error param

GregoryDosh:
Currently running into an issue where the try block above is throwing an error, but the error block is crashing the app because the error param is undefined on line 219. So this is simply putting the error param in the catch block so the error can emit properly.

### DIFF
--- a/src/providers/trayProvider.js
+++ b/src/providers/trayProvider.js
@@ -213,7 +213,7 @@ function updateImage(payload) {
                 ? nativeImage.createFromDataURL(payload) // electron v0.36+
                 : nativeImage.createFromDataUrl(payload) // electron v0.30
         tray.setImage(img)
-    } catch {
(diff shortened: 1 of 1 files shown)


mingjun97:
Thanks for contributing!

This changes has been done at commit f370bdc. So I'll close this PR. Feel free to reopen it if I miss something. :)

Btw, currently, master branch is for public release only, you may open PR to development branch next time.

GregoryDosh:
Nope this is great. Thanks for already catching that!

One small suggestion is putting into the contributors guide to make sure to check out the dev branch in your readme. https://github.com/ytmdesktop/ytmdesktop/#to-contribute I'm normally really good about catching those things but I was just in a hurry :-)

mingjun97:
Don't worry. :) 

We will address this in readme. Thanks for advising!

PR #624, opened 2021-02-15
Much more improved Miniplayer

flleeppyy:
_alright lets try this again without having a stroke with Github_

- Completely rewrote the convertToHuman() function (Very painful ;w; but much worth it)
- - This actually fixes an issue where the timestamp would be displayed wrong if it was longer than 60 minutes.
image

The rest is all miniplayer related.
- Add hover animations to control buttons
- Smooth solidify transition when liking/unliking

https://user-images.githubusercontent.com/18307183/107880462-bae8d100-6e93-11eb-9f5f-4f806780493c.mp4

### DIFF
--- a/main.js
+++ b/main.js
@@ -1262,6 +1262,9 @@ async function createWindow() {
                 }
             })
 
+            // Devtools
+            // miniplayer.openDevTools()
+
             mainWindow.hide()
         }
     }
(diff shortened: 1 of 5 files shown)


mingjun97:
Merged, thanks for your generous contribution!

Issue #249, opened 2020-07-13
YT Music app is very slow. Slowing down system, having to use web version.


DenneyLewis:
YT Music app is very taxing on my system. I'm not sure if there's something I'm doing wrong, but the app hogs a lot of resources, and causes my entire system to run slowly, whereas using the web version is just fine. App lags when hitting next song, has loading issues, and sometimes just blanks out all together.

PC Specs:
Windows 10
AMD Ryzen 5 3550H @ 2.10 GHZ, overclocked up to 3.70 GHZ
8GB Ram

Normally have no problems doing heavy activities. (I play World of Warcraft on high setting frequently with no issue) so i'm not exactly sure why YT music is taxing my system so badly.

alexaandru:
I came here to report the same. It's unusable unfortunately. I have `Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz` and `16GB` of RAM and use `Ubuntu 20.04 LTS`. Looks great though! Thank you! :)

adlerluiz:
thanks for reporting, are you using version 1.11.0?

alexaandru:
Yes, I am.

adlerluiz:
the problem of slowness already happens when opening, or while using the app?

mrmichaelbeck:
I have a similar issue.  I let the player play all day at work as I'm working.  After a day or two of this, the player becomes very slow.  It's a gradual build.  I'm on a dual-xeon machine with 40GB of RAM in Windows 10.  And I'm running 1.11.0.  Turning off the player and then starting it up again seems to speed things up again, but gradually it slows down again.

adlerluiz:
I believe that the use also happens through the youtube music site, since many people are complaining about it

https://www.reddit.com/r/YoutubeMusic/comments/hi9i9v/in_the_time_it_took_me_to_write_this_post_memory/

rhwendt:
I think this is related to https://github.com/ytmdesktop/ytmdesktop/issues/260#issue-659446882

adlerluiz:
Could you confirm that this is still happening with the new version?

mrmichaelbeck:
It's significantly reduced, but still occurs.

DenneyLewis:
I've been in this conversation since the beginning but haven't said
anything. From my side, everything is all good now, but I did just recently
buy a new gaming desktop so that very well could've fixed the issue, but I
did also download the new update. Either way, I'm not having any problem
now.

[...]

PR #677, opened 2021-03-25
Disable disk caching option

trinode:
Chromium / electron seem to write all streamed media to disk needlessly by default, this PR seeks to disable that.

By way of comparison these are stats from launching, searching and playing the first 3 mins of a video:-

Caching on:-

image

Caching Off :-

image

if you listen to music continuously during the work day this is over 4GB of unnecessary writes.

A similar flag was added to chrome to reduce battery usage caused by constant disk writes specifically for streaming media - so that should also be a pleasant side effect of this PR - though I won't claim to have proof.

### DIFF
--- a/main.js
+++ b/main.js
@@ -106,6 +106,8 @@ app.commandLine.appendSwitch('disable-features', 'MediaSessionService') //This k
 if (!app.isDefaultProtocolClient('ytmd', process.execPath)) {
     app.setAsDefaultProtocolClient('ytmd', process.execPath)
 }
+
+app.commandLine.appendSwitch('disable-http-cache')
     
 createCustomAppDir()
 


Venipa:
also fixes apple m1 disk mass caching issues

trinode:
Quick thought, would it be better to disable the cache by default and have the option be to enable it? If this app is silently unneccessarily wearing SSDs it'd be better to not do that by default but enable it if the user chooses to?

adlerluiz:
please switch to branch development to be accepted

trinode:
@adlerluiz I've updated the branch, but:

I'm having an issue now it's based on develop though with the i18n stuff just displaying the name instead of the value:-
image

I see there's another repo that also contains localisation stuff as well as it being here in this repo, have I missed something?

Also, it seems development has a husky job on commit to reformat files (it also complains about a extra closing div before I've made changes), so I've let that run but it's made it look like I've changed a lot of code in the settings.html file.

mingjun97:
It looks like the YTM still tries to re-fetch media files even the cache is enabled (by default). I think we could make it disabled by default (I mean don't create a entry at the settings panel). In my opnion, `enable cache` is somehow a confusing concept (Since it does not reduce the bandwidth/data utilization in this case). Lol

cyraid:
*Waiting for push* :D

trinode:
> _Waiting for push_ :D

Just to be clear, I have pushed my code, we're waiting for the pull / merge :)

cyraid:
> > _Waiting for push_ :D
> 
> Just to be clear, I have pushed my code, we're waiting for the pull / merge :)

Yes haha of course! I used the wrong word there, I meant merge.

mingjun97:
I'm going to merge this PR. Thanks for your great contribution!

Btw, wIll this PR close #699 #249 ?

trinode:
> 
> Btw, wIll this PR close #699 #249 ?

I'm not sure, it might be worth checking the original reporters of those issue if a build with this in helps, I didn't test memory usage just disk usage.

mingjun97:
image

Looks great for the memory leak issue as well. Let's run more test to see if it actually works.

Again, thanks for fixing this tricky problem. :)

unbekannt3DE:
I've downloaded the Windows CI build for testing but still have high RAM consumption on large playlists and long radio mixes. Also, don't know if it's due to RAM consumption but with long radio mixes the app gets slower and slower and sometimes skips several songs in queue when changing songs.
image

ddrobny:
@unbekannt3DE, Pardon me, but where do you find the CI build files? I see the actions for the different builds but don't know where to download the results.

unbekannt3DE:
@ddrobny on the commit merge from mingjun97 you can click view details and from there you can download the build artifacts 
image

ddrobny:
Thanks! Apparently I never paid attention to that corner! :-P

ddrobny:
> 
> I've downloaded the Windows CI build for testing but still have high RAM consumption on large playlists and long radio mixes. Also, don't know if it's due to RAM consumption but with long radio mixes the app gets slower and slower and sometimes skips several songs in queue when changing songs.
> 

I can confirm. RAM usage gets really high (Win10). I kept the automatic mix after the end of a playlist playing for a day and the RAM usage is nuts. Also, occasionally songs started to get stuck and skipped for unclear reason.
YTMD

PR #679, opened 2021-03-29
Fix infoPlayerProvider artist and album info

### LINKED ISSUE #120: Bug: App is not sending album/album artist to Last.fm
I noticed that a track I had listened to this week on my phone with YTM + Scroball app, and GPM desktop player on another device, was being scrobbled differently when I tried out this YTM desktop player. I have a last.fm subscription so I can go in and edit scrobbled metadata and found that the tracks from from YTM desktop were missing album and album artist data. Can that be enabled in the next update? I edited those scrobbles already but can do it again for an example if needed!

raunovillberg:
Fixes #120 
On Windows 10 and with both 1.13 & building from latest development branch I noticed that tracks were being scrobbled to last.fm with incorrect info.

For a specific example,: when listening to the album "Haram" by Armand Hammer & The Alchemist, "Haram" was set as the artist name and the album name was not parsed at all (so it was defaulting to "_")

With these changes, the album & artist info is parsed from the bar in the middle at the bottom while a track is playing:
YouTube_Music_Desktop_App_m5g4mVp0Eh
YouTube_Music_Desktop_App_Sod34Iplvo

Two related things in one PR - I'll happily split it into two if that'd be preferred :)

### DIFF
--- a/src/providers/infoPlayerProvider.js
+++ b/src/providers/infoPlayerProvider.js
@@ -223,17 +223,14 @@ function getAlbum(webContents) {
         .executeJavaScript(
             `
             var album = '';
-            var player_bar = document.getElementsByClassName("byline ytmusic-player-bar")[0].children;
-            var arr_player_bar = Array.from(player_bar);
+            var artist_album_and_year = document.getElementsByClassName('subtitle ytmusic-player-bar')[0].textContent;
+            var split_by_dot = artist_album_and_year.split(" • ");
             
-            arr_player_bar.forEach( function(data, index) {
-            
-            if (data.getAttribute('href') != null && data.getAttribute('href').includes('browse')) {
-                album = data.innerText;
-            }
-            } )
-            
-            album
+            // Ensure that we actually have album data set
+            if (split_by_dot.length > 1) {
+                album = split_by_dot[1];
+            }            
+            album;
             `
         )
         .then((album) => {
--- a/src/providers/infoPlayerProvider.js
+++ b/src/providers/infoPlayerProvider.js
@@ -201,14 +201,10 @@ function getAuthor(webContents) {
     webContents
         .executeJavaScript(
             `
-            var bar = document.getElementsByClassName('subtitle ytmusic-player-bar')[0];
-                        
-            if (bar.getElementsByClassName('yt-simple-endpoint yt-formatted-string')[0]) {
-            title = bar.getElementsByClassName('yt-simple-endpoint yt-formatted-string')[0].textContent;
-            } else if (bar.getElementsByClassName('byline ytmusic-player-bar')[0]) {
-            title = bar.getElementsByClassName('byline ytmusic-player-bar')[0].textContent;
-            }
-            title;
+            var title = '';
+            var artist_album_and_year = document.getElementsByClassName('subtitle ytmusic-player-bar')[0].textContent;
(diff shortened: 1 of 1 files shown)


kabili207:
I was about to start a pull request for this exact issue. I'm glad I'm not the only one who was looking into this.

Alipoodle:
This PR is also working fine with stuff for Songs which Don't have Artists or Albums and Both missing Albums & Artists.

Shouldn't `Album` be trimmed as well?

raunovillberg:
> Shouldn't `Album` be trimmed as well?

I'm not sure, perhaps neither of them should be actually trimmed.
`split(" • ")` already removes one whitespace on either side of the bullet... and perhaps there's a valid usecase for an artist to have a whitespace at the end or an album to have a whitespace at the start (though I'd guess most of the time that'd just be typos)

mingjun97:
image
For some songs that don't have `Albums` and `Years` information, the probably result may display like this which is `BELLA PING MUSIC CHANNEL 2 • 170K views • 1.7K likes`. Kinda tricky. In this case, it looks like your commit will parse the count of views as the album name or I'm wrong? Kinda tricky, lol.

kabili207:
That's going to be a lot harder to handle since those are really YouTube videos. The only thing we're going to be able to pull from those are the title. It looks like we can detect videos if we look for the presence of the `video-mode_` attribute on either the `ytmusic-player-page` or `ytmusic-player` elements. That attribute doesn't exist for non-video songs.

Screenshot_2021-05-13_11-29-01

blackouty:
> Fixes #120
> On Windows 10 and with both 1.13 & building from latest development branch I noticed that tracks were being scrobbled to last.fm with incorrect info.
> 
> For a specific example,: when listening to the album "Haram" by Armand Hammer & The Alchemist, "Haram" was set as the artist name and the album name was not parsed at all (so it was defaulting to "_")
> 
> With these changes, the album & artist info is parsed from the bar in the middle at the bottom while a track is playing:
> YouTube_Music_Desktop_App_m5g4mVp0Eh
> YouTube_Music_Desktop_App_Sod34Iplvo
> 
> Two related things in one PR - I'll happily split it into two if that'd be preferred :)


How to install?

Spaeck:
@blackouty You'll have to build from source yourself, using node.js and electron-builder. I compiled the app for my self, [here (mega.nz)](https://mega.nz/file/KZp2yRzR#KTj0_LbzuqAV6de82nbKlRHpNmcal4Jgt0STCfB8kjk) is the installer in case you want to use it. Only changed the file from this PR.

Spaeck:
> How to install?

@blackouty You'll have to build from source yourself, using node.js and electron-builder. I compiled the app for my self, [here (mega.nz)](https://mega.nz/file/KZp2yRzR#KTj0_LbzuqAV6de82nbKlRHpNmcal4Jgt0STCfB8kjk) is the installer in case you want to use it. Only changed the file from this PR.

zejjnt:
Could it be fixed by regex?
It is in no way foolproof but a video would give something like _artist_ • [\d\.]+(K|M)? views • [\d\.]+(K|M)? likes\\n
while a regular track would just be _artist_ • [\d]{4}\\n

(I am bad at regex, I do not code JS, I am tired and kind of drunk but I am still pretty sure my point comes across)

Alipoodle:

PR #728, opened 2021-05-24
:bug: don't refresh icon if no progress bar

ranvel:
- macOS: Dock icon constantly refreshes even if there is no activity, this addresses that issue as verified with Quartz Debug


mingjun97:
Merged. Thanks for your contribution!

PR #692, opened 2021-04-12
Add MPRIS volume support and Fix setSeekbar()

Pauloo27:
MPRIS volume, both "get" and "set"

PR #755, opened 2021-06-29
Add support to volume in MPRIS

Pauloo27:
MPRIS volume, both "get" and "set".

(basically #692 but now done right)


adlerluiz:
Thank you!

Issue #740, opened 2021-06-11
"Continue where you left off" does not work


pdario:
- [x] I understand that **YTMDesktop have NO affiliation with Google or YouTube**.
- [x] I verified that there is no open issue for the same subject.

**Describe the bug**
"Continue where you left off" used to work on my old PC; now I changed it and installed YTMD back, but now the "continue where you left off" feature does not work: I loose the playlist I was listening to and I have to manually start a new one.
I tried cycling the toggle off and then on again.

**To Reproduce**
1. Start a Playlist or shuffle
2. Close YTMD
3. Restart YTMD
4. Nothing plays

**Expected behavior**
"Continue where you left off"

**Screenshots**
N/A

**Environment (please complete the following information):**
 * YTMDesktop version: 1.13.0
 * OS: Windows
 * OS version: 10 21H1
 * Arch: x64
* Installation way: installation package

KegNot:
Same here on W10 20H2.
I'm new to YTMD so I can't tell about previous behavior.
Did all the things pdario did with no effect.

Before discovering this github I sent an e-mail to Adler. Hope someone will respond.

KegNot:
I found pre-release v1.14.0 and tried that. Nope. "Continue where you left off" doesn't work with that version either.

KegNot:
Got a reply from Adler Luiz. It has to with the play bar. Whether it has focus at the moment the app is shut down. His literal reaction: "Well it's a 'little problem' in the way youtube music works, there's only way to know which playlist or song you're currently on, if you have the bar open, so the 'resume where you left off' mode will work."

This "state" will not remember because the bar is "closed":
![image](https://user-images.githubusercontent.com/59098871/122671580-add44980-d1c7-11eb-9f50-2e1093efbd62.png)

This "state" will remember because the bar is "open":
![image](https://user-images.githubusercontent.com/59098871/122671645-fdb31080-d1c7-11eb-8459-5d2a6c6408f4.png)

Click the triangle at the bottom right (or any "empty" part of the bar) to open or close the bar.

mingjun97:
The way YTMD dealing with `continue where you left off` is to record the URL you currently viewing. With bar on, you can get URL like this.
![image](https://user-images.githubusercontent.com/15214491/123014551-5bd42500-d37b-11eb-935c-3321313635e8.png)

With bar off, your URL will be like this.
[...]

PR #765, opened 2021-07-09
🐛 Fix continue where you left off

NovusTheory:
Related to #740

Fixes the continue where you left off setting where it couldn't correctly save the queue. YouTube Music exposes an element that provides the session link for the currently playing song and playlist together so instead of using the window URL as it previously was, it will use that. Do note I made it so if session link isn't found it will default to the behavior it was doing before which is saving the current window URL.

Known issues:
-  Radio and autoplay queues aren't saved. This is upstream to YouTube Music and cannot be fixed.

### DIFF
--- a/main.js
+++ b/main.js
@@ -612,6 +612,13 @@ async function createWindow() {
                                 )
                             }
                         })
+                        .catch(() => {
+                            // JavaScript errored, assume no session link found and default to current url
+                            settingsProvider.set(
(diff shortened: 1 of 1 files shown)


NovusTheory:
CI note: Failed due to outdated lock file, I feel this is not in scope for this PR to fix. This PR does correctly compile

PR #777, opened 2021-07-31
Updated packages and migrated the code to Electron to 13

SnakePin:
Updated most package major versions.
I've checked most of the code and there were no breaking changes that I could find.
Changed the places where the now deprecated Electron functions were used such as the `electron.remote` and `new-window` event handler.
`electron-is-dev` was not updated as it introduced breaking changes.
Fixed the event emitter limit warning that showed up in the console which was caused by the settings provider subscriptions.


adlerluiz:
Thank you very much for your help in uploading the version, I will finish making other changes and test your change to make sure nothing is missing.v

PR #805, opened 2021-08-29
:bug: setVolume function

nblath:
Using an event keeps the UI in sync and allows volume 0

### DIFF
--- a/src/providers/infoPlayerProvider.js
+++ b/src/providers/infoPlayerProvider.js
@@ -442,8 +442,8 @@ function setVolume(webContents, time) {
         .executeJavaScript(
             `
         var slider = document.querySelector('#volume-slider');
-        slider.value = ${time};
-        document.querySelector('.video-stream').volume = ${time / 100}
+        slider.step = ${time} - slider.value;
+        slider.dispatchEvent(new KeyboardEvent("keydown", {"key":"ArrowRight"}));
         `
         )
         .then()


Nerogar:
Hi. I tried to fix the same issue a while ago and came up with a similar solution. YTM always rounds volume values to the nearest integer (or maybe it just truncates decimal places, not completly sure). This leads to the following behavior which I also couldn't solve:

If the option to replace the volume slider with dB values is enabled, the volume always contains decimal places. This is especially useful for lower volumes <5%. If the user uses the volume keys to lower the volume below 1% with this fix, the volume is set to 0% ( => muted). Raising the volume is very buggy now.

So we have two separate problems:
1. Fine control over the volume in the 0%-2% range is not possible anymore.
2. Raising the volume from 0% is very buggy.

I already thought about possible solutions for that, but none of them seem to work properly. The previous solution directly set the volume of the video html element. That is not possible anymore, because YTM now regularly updates this element with some internal state that is itself updated by the ui (but not direcly linked to the ui).

I still think this solution is better than the current implementation on the development branch, which is completely broken right now. But maybe we could think about possible solutions at least for the second problem before merging this PR.

nblath:
Hi. It seems I can get the volume-slider to accept some decimal values, but this needs to be investigated further as it seems a bit inconstant. I also managed to figure out a better solution for the setVolume function that is closer to that of the prior working solution. It seems that after changing the value attribute of the of the volume-slider you can simplify dispatch a change event to get the player to accept the new value and update its internal state.

I've introduced a quick workaround for the problems that occur with the dB setting when the volume is < 6.0%. Below that level it has issues with getting the player to accept such a small change in value. So far I have not been able to get anything constituently working with increments < 1.0%. As far as I can tell, this is not an issue caused by rounding or truncation, but possible the way the slider it self works.

adlerluiz:
yes it seems to be a slider behavior, but we have to see the impact of this change, but i will fix some other things locally and test this modification locally.

Alipoodle:

PR #823, opened 2021-09-24
🌐 Update it.json

albanobattistella:


Alipoodle:
Considering both of these are Brand/Service names rather than general language, I don't think these translations are actually appropriate.🤔

I wouldn't Translate YouTube as manually `You` `Tube` like VoiTubo

albanobattistella:
> Considering both of these are Brand/Service names rather than general language, I don't think these translations are actually appropriate.thinking
> 
> I wouldn't Translate YouTube as manually `You` `Tube` like VoiTubo

Hi, ok, I seen.

PR #839, opened 2021-10-11
Move to Xosms

NovusTheory:
This PR will move all the OS media service providers from using nodert (and ultimately my fixed package for it) & mpris-service into a new package that I've created called Xosms which is a cross platform media service package which as of right now supports Windows & Linux with macOS planned to be supported. Xosms is conditionally compiled for each platform and will noop itself if the platform is unsupported.


dm3ch:
I'm interested what is the actual status of this PR?

Alipoodle:
Progress still needs to be made on Xosms' side really 😅
Novus wants to ideally have this to fully replace all Media Services we have and not have one for each platform.
Currently MacOS is lacking functionality within the Xosms.

dm3ch:
@Alipoodle Thank you for your answer.

As far as I understand right now even without XoSms ytmdesktop lacks MacOS media integration support, so it seems that even without that part it doesn't make things worth.

P.S. Also there's a PR implementing MacOS support for XoSMS but it needs a review - https://github.com/NovusTheory/xosms/pull/11

dm3ch:
I'm interested why this PR was closed.

Right now xosms supports all 3 platforms.
@NovusTheory Do you have plans to finish work on this?

I'm asking cause it's painful to use this app on MacOS without any media integration

sfllaw:
There’s a PR that replaces this one: #1359.

PR #844, opened 2021-10-14
✨ Added more features to the existing restful API ✅

DamienStaebler:
added skipAd function to API


KillerBOSS2019:
Do I just delete `package-lock.json` and then run `yarn install`?

Issue #793, opened 2021-08-21
Rainmeter WebNowPlaying SetVolume always decreases the Volume


StefanTieber:
- [x] I understand that **YTMDesktop has NO affiliation with Google or YouTube**.
- [x] I verified that there is no open issue for the same subject.

**Describe the bug**
Using the WebNowPlaying Plugin for Rainmeter you can send a SetVolume Command (or Bang as it is called in Rainmeter) to YTMD. This always decreases the volume no matter the argument.

**To Reproduce**
1. Load the 'WebNowPlayingExample' skin by tjhrulz in Rainmeter
2. Open YTMD and begin playing any song
3. Click on the volume increase button on the Rainmeter skin
4. See the volume of YTMD go down

**Expected behavior**
Volume increases

**Environment (please complete the following information):**
 * YTMDesktop version: ***v1.14.0***
 * OS: ***Windows10***

Alipoodle:
Looked into this.
Appears the documentation states 2 different ways of changing volume:

> `SetVolume ###`
> Where ### is a Integer beween 0-100, add + or - in front to set the volume relatively.
For the WebNowPlayingExample they use the `+5` option which is currently not implemented within our Integration for Rainmeter.

This is a very simple fix, so I'll look into it when I'm getting on to some PRs

Alipoodle:
Closing issue due to no longer supportinng YouTube Music Desktop v1 + Release of YouTube Music v2.
Please re-open this issue if this is still occuring in the latest build of the application, however as a re-write has been done on the entire project most issues will be fixed, no longer a feature or re-worked.

PR #847, opened 2021-10-16
🐛 Fix WebNowPlaying volume always going down.

Alipoodle:
This is a bit of a short and simple fix for this issue.
We fail to actually set the value exactly within YTM but this fixes the core issue of the volume not changing when selecting up/down.

Fixes: #793


niruvibes:
👀

PR #896, opened 2022-01-28
feat(discord): Use track cover in discord RPC

germanoeich:
Discord announced today that rich presence now supports dynamic URL images. With this, it's possible to use the album art 🎉 

Looks like this:
image


NovusTheory:
This is already in our development branch

PR #899, opened 2022-02-03
💎 Remote Control PORT + Window OOB fixed

eltharynd:
Added a port input to the settings page so user can chose what port to use for the server.

Also fixed a minor bug ~~where development on ultrawide screens~~ _that_ would push the window out of bounds when closed incorrectly


eltharynd:
In regards to this:

> Also fixed a minor bug where development on ultrawide screens would push the window out of bounds when closed incorrectly

I actually dug deeper into the issue... turns out:
- the window was created **before** those checks were made (so changes weren't applied)
- the math to validate the window position/size in the viewport was wrong (and since they were saved onClose it would f**k up whenever you force closed from terminal)

So i rewrote the code itself to make it actually work

main.js
```diff
@@ -1884,33 +1898,41 @@ else {
     })
 
     app.whenReady().then(async () => {
-        checkWindowPosition(
+        var adjustedWindow = await checkWindowPosition(
```
Alipoodle, line 1901: Why are we awaiting for this now?

main.js
```diff
@@ -1884,33 +1898,41 @@ else {
...
             width: 700,
             height: 800,
         })
-            .then((visiblePosition) => {
-                console.log(visiblePosition)
-                settingsProvider.set('lyrics-position', visiblePosition)
+            .then((adjustedWindow) => {
+                console.log(adjustedWindow)
+                settingsProvider.set('lyrics-position', adjustedWindow.position)
+                settingsProvider.set('lyrics-size', adjustedWindow.size)
```
Alipoodle, line 1918: Why are you saving Lyric Size, if you then aren't using it on Lines 1912-1913

src/pages/settings/settings.html
```diff
@@ -870,6 +870,23 @@
...
+                                            <span
+                                                class="settings-item-last-child"
+                                            ></span>
+                                            <span
+                                                i18n="i18n_LABEL_SETTINGS_TAB_GENERAL_COMPANION_SERVER_PORT"
+                                            ></span>
+                                        </td>
+                                        <td>
+                                            <input
+                                                type="text"
```
Alipoodle, line 884: Why is this Text and not a Number?
This would greatly reduce the amount of checks we would need to do and would guarantee a number at least.

Alipoodle (commented):
This is an initial review before actually running the code.
Just wanted to check up on a few things before running fully.

main.js
eltharynd: Because this needs to be completed before running

``` javascript
await createWindow()
```

at line 1939.
That was the whole reason the changes from `checkWindowPosition` weren't applied, and the reason why they were applied (even if wrong) if you halted the program ungracefully.

This resulted in the window opening outside of the monitor bounds.

main.js
eltharynd: It was already like that and didn't wanna change it in case whomever did it intentionally wanted to try and run it at the size in those lines.

I merely updated the code to receive a

``` 
adjustedWindow {
    position
    size
}
```

object.

Same for the `checkWindowPosition` right below this one

src/pages/settings/settings.html
eltharynd: Even if you put type="number" that does not prevent the user from inserting text.

For this reason I opted for a type="text" simply because it does not display the arrows to the right, which I would've hidden for stylistic reasons.

The port gets validated in the logic to be a number and in range

eltharynd:
closing cause it's inlcuded in #900

PR #900, opened 2022-02-04
🐛 socket.io update

eltharynd:
I updated the socket.io package to latest version.

Additionally, when not used from localhost there was no guarantee the password would've worked.

This is due to the fact that the WebSocket standard prevents from sending custom headers, and that's what was used in the app.

You could've solved this by using a standard header that's supported, but the better version is to just update socket.io as the new versions have a built in auth system.

I kept the old method in there for retro-compatibility.


Alipoodle:
9864 can't be used currently as this is used by the Genius Callback URL, are you able to update this to handle the stuff for that. 😅

eltharynd:
@Alipoodle i can't find any other usage of the socket? could you point me to the code that's not working? i have been using my own build of this since i made it and it works fine but i'm not sure what genius is

eltharynd:
@Alipoodle nvm i found out what genius is...

i just defaulted the companion server port to a different one and included it in the validation

Alipoodle:

PR #924, opened 2022-03-13
:bug: fix hang on npx

SomeAspy:
I wasn't able to build this (I still can't, but it won't hang with this fix) and found it kept hanging on the npx command. following https://github.com/electron-userland/electron-builder/issues/6411 I replaced npx npm exec.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,13 +7,13 @@
         "start": "electron .",
         "dev": "NODE_ENV=development nodemon --exec yarn start --dev --inspect",
         "rebuild": "electron-rebuild -f",
-        "postinstall": "npx electron-builder install-app-deps",
-        "build:win": "npx electron-builder --win --config electron-builder-deploy64.yml",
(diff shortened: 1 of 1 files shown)


Alipoodle:
Some work is going to need to be done regarding fixing our actions with this PR.
Build process is currently using NPM 6 which causes this to fail as `npm exec` is a npm 7 release...

I wanna do some investigation.
Can I ask why you've deleted this so soon?

SomeAspy:
Sorry I saw the fail, and realized it didn't work

Alipoodle:
There's a few reasons why this could hang for you.
I'm assuming you are running Windows 10/11.
You need Visual Studio 2015 installed for the NodeRT libraries to compile correctly as they are hard coded to this version of Visual Studio. We want to entirely remove these but currently we're waiting on another library to be ready to be installed and for it to work a lot better.
Otherwise that is the main issue I have found regarding installing for people.

SomeAspy:
I tried building it using npm, pnpm, and yarn, on wsl and windows 11 I has VS 2017 and 2022 installed so let me try 2015

SomeAspy:
@Alipoodle I installed vs 2015 and I still have the issue where it freezes

SomeAspy:
nevermind, I got it, thanks!

PR #942, opened 2022-04-13
✨ Alternate Miniplayer Stream Widget

PHNX-Arcanus:
ONCE MORE, WITH FEELING

Many "Now Playing" music widgets for use in OBS/XSplit usually require multiple third party services or software programs, and more often than not only have support for Spotify or web-hosted music streaming. Additionally, due to the information passing between multiple APIs it is common for the information displayed to not be accurate or sufficient. This feature is intended to be a baked-in solution for the YouTube Music Desktop Application, by shuffling around the existing elements of the Miniplayer for more optimized screen space and information display. I have added a toggle in the Miniplayer settings menu, which will load alternate HTML/CSS files that play nicely with these streaming tools, as well as a tooltip on the toggle to indicate to the user settings that this feature will ignore.

Alright, take two - Started from scratch, hopefully this should be better, localization PR to follow.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1272,72 +1272,68 @@ async function createWindow() {
     async function windowMiniplayer() {
         if (miniplayer) miniplayer.show()
         else {
-            if (settingsProvider.get('settings-miniplayer-stream-config')) {
-                miniplayer = new BrowserWindow({
-                    title: __.trans('LABEL_MINIPLAYER'),
-                    icon: iconDefault,
-                    modal: false,
-                    frame: false,
-                    center: false,
+            var miniplayerConfig = {
+                title: __.trans('LABEL_MINIPLAYER'),
+                icon: iconDefault,
+                modal: false,
+                frame: false,
+                center: false,
 
-                    width: 500,
-                    height: 100,
-                    resizable: false,
-                    skipTaskbar: false,
+                resizable: settingsProvider.get(
+                    'settings-miniplayer-resizable'
+                ),
+                skipTaskbar: !settingsProvider.get(
+                    'settings-miniplayer-show-task'
+                ),
+                alwaysOnTop: settingsProvider.get(
+                    'settings-miniplayer-always-top'
+                ),
 
-                    alwaysOnTop: settingsProvider.get(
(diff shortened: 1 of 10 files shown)


Alipoodle:
This is looking really promising.
Will test this weekend!

Alipoodle:
Looking into this today, and done a bit of work on it to make it responsive. 🥰
Probably will not have it save the Size for the time being, but might be something to add later.

DiscordCanary_2022-04-18_14-35-44
Need to fix it's minimum size though. 😅

PHNX-Arcanus:
Looks great!  Thanks for giving this a gander.

Alipoodle:
So I'm happy with what I've got it with now. My only question I want to ask is, why was the `taskbar visibly` option ignored for this? I feel this should be an option which people should still be able to choose as this allows for more customisability with the existing settings that exist? 

I also am going to tweak it so that the Additional note is a Translation string as well.
Will push it into your Repo and Merge the PR later today.

> Changes the layout of the miniplayer, optimizing screen space for easy use with live streaming. The following settings will be ignored: miniplayer size, resizability, and taskbar visibility.

PHNX-Arcanus:
I found that the only way OBS/Xsplit can hook into the window and display on-stream is when the taskbar is visible.

Alipoodle:
Please note that this may still be used by people as a non-stream feature as some people might prefer the wider nature of it.
I will quickly try poking into OBS [as that what I use and it's free] and seeing if this fails to work.

Alipoodle:
Any objects to the local update I have done?
https://github.com/ytmdesktop/ytmdesktop-locales/pull/20

Otherwise I will happily push both of them into development 😄

PHNX-Arcanus:
Should be all set!

PHNX-Arcanus:
Awesome!  So is this something that will make its way into the master build?

Alipoodle:
It will likely make it into the nightly, which I plan to do soon as I go through some of the mature PRs.

Alipoodle:
The controls still exist on the UI due to comparability with the miniplayer file but they're hidden currently for the view so this is definitely possible.

PR #943, opened 2022-04-14
:sparkles: Change playback speed

wepiotrek:
### Playback speed feature
Feature that allows to play music with different speed (known from normal YouTube, Netflix etc.).

For my experience music sounds better with speed x1.3-1.4 (but you can choose with your preferences).
Fig1

When you drag mouse on the speed icon you can use slider to change speed ( from x0.3 to x2.0 - my suggestion):

Fig2

Edit: Another use case: podcasts & audiobooks (if someone speaks too slowly or too fast).

### DIFF
--- a/main.js
+++ b/main.js
@@ -633,6 +633,7 @@ async function createWindow() {
 
                 writeLog({ type: 'info', data: `Listen: ${title} - ${author}` })
                 discordRPC.setActivity(getAll())
+                infoPlayerProvider.initializeSpeed()
             }
 
             /**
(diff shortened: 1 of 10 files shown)


A-Matt:
Is it possible to get this as a item in the UI Actions to Hide and Show this option?
image

Alipoodle:
I agree with A-Matt and probably worth giving a small bit of transition on the showing and hiding of the element as well. 😅
it's a bit to abrupt to show it.

Alipoodle:
I was quickly trying to get it to nicely Show/Hide but due to it being a display none/block show and hiding, I was unable to do this in the short time I want to put into looking at PRs and "fixing" them.

This is something which I feel would be really IDEAL to include to make it look a bit more clean like what YTM would have.

Alipoodle:

PR #949, opened 2022-04-21
Background Color painting + discord rich presence 

reHamad:
+ Instead of switching colors from one song to the another, set it to **Fade** to the next color since that will be better looking and without any pops from one color to the another.

note: by pops I mean the color goes to black and then the next color pops out.

+ Discord rich presence that displays the album cover instead of the YT Logo should be out for discord users!


Alipoodle:
Please open this as an issue and not a PR request
Additionally Album Art as Discord Rich presences is a feature in development which is coming in next Nightly / Release.

PR #954, opened 2022-04-28
bug fixes

psycodeliccircus:
fixed the pt-BR translation for the system.

PR #955, opened 2022-04-28
Fix Bugs

psycodeliccircus:
Bug fixes and updates


Alipoodle:
The addition of the **Português (Brasil)** Language in the Settings is something which is indeed missing so can you add that to the **Development** Branch of ytmdesktop instead of the **Master**. (I will probably do it myself if left undone for a while)

AS stated within the previous issue #954
We use [ytmdesktop/ytmdesktop-locales](https://github.com/ytmdesktop/ytmdesktop-locales/) for the Locales now, and these already fixed.
Additionally we are using the `Development` Branch as a the main section of where progress is added to, and the locale files for 


Are the relevant files over there if you wish to add locales to that.
https://github.com/ytmdesktop/ytmdesktop-locales/blob/master/locales/pt-BR.json
https://github.com/ytmdesktop/ytmdesktop-locales/blob/master/locales/pt.json

PR #960, opened 2022-05-03
Update infoPlayerProvider.js - Fix images for queue not in api

Gratenes:
Change the area from which the que list is grabbed. This new path will resolve the correct images once they have been previewed by the user. 
*A better way for this needs to be done because i don't really like the use of the "[1]"*

### DIFF
--- a/src/providers/infoPlayerProvider.js
+++ b/src/providers/infoPlayerProvider.js
@@ -307,7 +307,7 @@ function getQueue(webContents) {
     webContents
         .executeJavaScript(
             `
-            var element = document.querySelector('ytmusic-player-queue #contents')
+            var element = document.querySelectorAll('ytmusic-player-queue #contents')[1]
             var children = element.children
 
             var arrChildren = Array.from(children)


Alipoodle:
I've re-tweaked this locally to at least check if there is 2 elements available.
But this doesn't seem to fully fix the issue as later down the queue it will return errors still.

image

this Might be something worth including for-now but needs to probably to be fixed as to why the later things are still breaking.

Gratenes:
> I've re-tweaked this locally to at least check if there is 2 elements available. But this doesn't seem to fully fix the issue as later down the queue it will return errors still.
> 
> image
> 
> this Might be something worth including for-now but needs to probably to be fixed as to why the later things are still breaking.

This is a problem with the dynamic loading, the user would have to scroll to the bottom and load them all.

Alipoodle:
I had scrolled through and visually showed all of those album arts and it didn't update these 😅
that's what I flagged this.

Gratenes:
main youtube changes there stuff daily 🤣

Alipoodle:
Yea it's annoying... that's some of the struggles with this sort of thing of wrapping YTM Desktop.
I'll do some more investigations into this as I need to fix my "enlarge" Album art I have.

`.side-panel.modular` then loop: `ytmusic-player-queue-item.style-scope.ytmusic-player-queue .left-items .thumbnail img` 
Seems to work, but needs items loading to fix. 
Might have to see if we can manually scroll the playlist to get the items in it. ETC.

PR #974, opened 2022-06-04
Discord presence mimic Spotify

Gratenes:
There is a correct and easy way to get something to work...

Discord_zBjmPkhhDz
electron_BZ6fymsSlq


src/utils/defaultSettings.js
```diff
@@ -71,6 +71,7 @@ if (!accelerators['miniplayer-open-close']) {
 settingsProvider.setInitialValue('has-updated', false)
 
 settingsProvider.setInitialValue('discord-presence-settings', {
+    default: true,
```
Alipoodle, line 74: This needs to be modified to check if there is already a `discord-presences-settings.default` as the `setInitialValue` will not add it for existing YTM Desktop Users.

*Additionally I recommend setting this to an "alternativeLayout" style name instead of Default as the Default should be default, and people should optionally opt into a different style and know what they're opting into.*

src/pages/settings/sub/discord/discord_settings.html
```diff
@@ -101,6 +101,24 @@
...
+                                                <label>
+                                                    <input
+                                                        id="settings-discord-use-default"
+                                                        type="checkbox"
+                                                    />
+                                                    <span class="lever"></span>
+                                                </label>
+                                            </div>
+                                        </td>
+                                    </tr>
```
Alipoodle, line 121: I feel that the wording for the Setting should rather be "Use alternative design" instead of use default and flip the switch on how the Setting functions.
People should automatically be opted into the Default value and be enabling the setting knowing what they're changing the format of the Rich Presences.

Additional might be worth doing a similar thing to what was implemented for the alternative design for Mini-player and using a small tooltip at the end of setting label to describe a bit of what is happening regarding the Alternative Layout.

src/locales/scaffold.json
```diff
@@ -23,6 +23,7 @@
     "LABEL_SETTINGS_DISCORD_SHOW_IDLE": "Show track when playback is paused",
     "LABEL_SETTINGS_DISCORD_SHOW_TIME": "Show track time",
     "LABEL_SETTINGS_DISCORD_SHOW_TITLE": "Show title",
+    "LABEL_SETTINGS_DISCORD_USE_DEFAULT": "Use default",
```
Alipoodle, line 26: "Use alternative layout" - sounds better or word it how you wish.
as mentioned by my review in the `discord_settings.html`

src/locales/en.json
```diff
@@ -24,6 +24,7 @@
     "LABEL_SETTINGS_DISCORD_SHOW_IDLE": "Show track when playback is paused",
     "LABEL_SETTINGS_DISCORD_SHOW_TIME": "Show track time",
     "LABEL_SETTINGS_DISCORD_SHOW_TITLE": "Show title",
+    "LABEL_SETTINGS_DISCORD_USE_DEFAULT": "Use default",
```
Alipoodle, line 27: "Use alternative layout" - sounds better or word it how you wish.
as mentioned by my review in the `discord_settings.html`

src/providers/discordRpcProvider.js
```diff
@@ -41,70 +42,148 @@ async function setActivity(info) {
...
+    if (
+        (!discordSettings.hideIdle
+    )
+        await client.clearActivity()
+    else
+        await client?.request('SET_ACTIVITY', {
+            pid: process.pid,
+            activity: discordActivity,
+        })
+}, 5000)
```
Alipoodle, line 185: This should be moved to a new place so that it's not applied to ALL users, remember that this file will still get ran if the user has Discord RPC disabled.

Additionally I feel that there could be some clean up regarding setting the Discord Activity to null when hideIdle is off & we've already cleared the activity to not keep running it.

Alipoodle (commented):
Hey this is just an initial review before I've actually ran the work locally.
I feel that some of these changes are better for performance (including the interval) & the understanding of setting for the user.

Alipoodle:

Issue #912, opened 2022-02-19
Default volume keyboard shortcuts break basic Windows functionality


AlexHimself:
### Consent

- [X] I verified that there is no open issue for the same subject.
- [X] I understand that **YTMDesktop have NO affiliation with Google or YouTube**

### Current Behavior

The default **volume** keyboard shortcuts break standard Windows functionality. These specific shortcut keys are fundamental to Windows for basic text selection across all applications. It's similar to hijacking `Shift`+`<arrow key>`.

Generally, changing shortcut keys would be considered a feature request, but these specific shortcut keys are core to nearly every application that deals with text, I'd consider it a bug.

- `Ctrl`+`Shift`+`<Up Arrow>`
- `Ctrl`+`Shift`+`<Down Arrow>`

I've had issues using Microsoft Excel/Word/Office/Notepad/etc and it took a little while to figure out _why_ the most basic functions were broken.

### Expected Behavior

- `Ctrl`+`Shift`+`<Up Arrow>` - Should select all text in the upward direction until a break
- `Ctrl`+`Shift`+`<Down Arrow>` - Should select all text in the downward direction until a break

### Steps To Reproduce

1. Open a multi-line text application or multi-row Excel file
2. Place cursor in a location in the middle of the text or Excel file
3. Press `Ctrl`+`Shift`+`<Up/Down Arrow>`
4. [Bug] Cursor does not move and text is not selected

### YTMDesktop

v1.13.0

### OS

Windows

### OS Version

10

### Arch

x86

### Installation way

.exe

Alipoodle:
Shortcut can be changed and is possible for you to disable within the Settings Menu.
This does not seem like something which we need to change as this is just a Default value we set. 

I personally use CTRL + ALT + A (down) / Q (up)

AlexHimself:
I eventually found the shortcut, but it might take your average user a _long_ time to identify that application as the cause of apparent, broken system functionality.

I suppose it's up to you guys as the developers/maintainers. I'd just suggest a different default key binding.

I work in software in the business world and nearly every business-user I've dealt with is taught `Ctrl`+`Shift`+`<arrow>` because the arrow keys are intuitive and obvious for them.

AlexHimself:
**This is a bug**

Alipoodle:
@AlexHimself This is not a bug... The feature can be disabled by the user or recongifured to whatever they want, and is functioning 100% as intended [Pressing a keybind and an action happening]

[...]

PR #1071, opened 2023-01-13
Added button for enable disable global shortcut

khisby:
# Description
add button to enable/disable global shortcut.

# Context Background
when I use YTM Dekstop, I often press the wrong button that triggers the shortcut, and it's very annoying

# Support locales
"MEDIA_CONTROL_ENABLE_DISABLE": "Enable/Disable Shortcut Media Control",
https://github.com/ytmdesktop/ytmdesktop-locales/pull/23

# Screenshot Result
image
image

# How Has This Been Tested
i tested the local dev build and checked the shortcut work as expected

### DIFF
--- a/STORE_VARIABLES.md
+++ b/STORE_VARIABLES.md
@@ -38,6 +38,7 @@
 | `settings-volume-media-keys`               | Boolean | `true` or `false`            | `false`                      | Enable media volume keys for the playback volume                                     |
 | `settings-pause-on-suspend`                | Boolean | `true` or `false`            | `false`                      | Pause media playback on `suspend` event triggered                                    |
 | `settings-surround-sound`                  | Boolean | `true` or `false`            | `false`                      | Enable Chromium Flag "try-supported-channel-layouts" for surround sound/Speaker Fill |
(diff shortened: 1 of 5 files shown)


Alipoodle:
I don't see why this is needed when each individual keybind can be disabled?
and in turn can be done on every command.

We plan to remove all default keybinds in a future up due to this issue and it causing issues.
#912 has more details.

image

Example I only have the Volume on/off.
image

khisby:
> I don't see why this is needed when each individual keybind can be disabled? and in turn can be done on every command.
> 
> We plan to remove all default keybinds in a future up due to this issue and it causing issues. #912 has more details.
> 
> image
> 
> Example I only have the Volume on/off. image

I think it will be very easy if we can disable all shortcuts with just one button and without removing the previous key binding configuration.

Ithink it's not right to remove the default keybinds. you can remap default keybinds or **you can use this my PR and set it to default turn off and user can turn on if they needed**.

and the issue https://github.com/ytmdesktop/ytmdesktop/issues/912 can be solve.

khisby:
Hi @Alipoodle i've add some commit to fix shortcut issue and it to make sure work as expected

Alipoodle:

PR #1073, opened 2023-01-15
Update major dependencies. Now using Electron 22 and other required dependencies.

guywmustang:
Update the notification title for windows.

Added windows readme for how to setup and run as it is quite hard to setup.

Major version bump to 1.15.0.


guywmustang (commented):
My first github pull request so if I did something wrong let me know.

I started with the 'development' branch as my baseline then updated everything else bringing this package up to date.  It was quite a lot more effort than I initially though (especially given I've never used Electron before).

Issue #866, opened 2021-11-10
Can't sign in, Browser or app may not be secure


dancojocaru2000:
### Consent

- [X] I verified that there is no open issue for the same subject.
- [X] I understand that **YTMDesktop have NO affiliation with Google or YouTube**

### Current Behavior

Google doesn't allow me to sign in:

![Couldn't sign you in Google page screenshot](https://user-images.githubusercontent.com/9626761/141196557-9b1d6984-1f6d-42d1-b560-8c055b1d4e89.png)

### Expected Behavior

Sign in.

### Steps To Reproduce

Download app, click sign in

### YTMDesktop

v1.13.0

### OS

Fedora

### OS Version

34

### Arch

x64

### Installation way

.AppImage

Alipoodle:
On Windows machine here, but cannot seem to reproduce this.
We get around this issue with a different useragent than Electron Natively provides (to circumvent this)

Can I please confirm that you are indeed on 1.13?

nikobonomi:
Same problem here,

the login is stuck on the 2-step verification phase

I'm on ubuntu 20.04, YTMDesktop installed trough snap

here is the snap info:
```
name:      youtube-music-desktop-app
summary:   YouTube Music Desktop
publisher: Adler (adlerluiz)
store-url: https://snapcraft.io/youtube-music-desktop-app
license:   unset
description: |
  Listen YouTube Music without need open the browser
commands:
  - youtube-music-desktop-app
snap-id:      S0db4gRY0r3Galb1B51Bt5h7zKbwqNj8
tracking:     latest/stable
refresh-date: today at 10:07 CET
channels:
  latest/stable:    1.13.0 2020-09-07 (6) 76MB -
  latest/candidate: ↑                          
  latest/beta:      ↑                          
  latest/edge:      ↑                          
installed:          1.13.0            (6) 76MB -
```


![Screenshot from 2021-11-15 10-13-52](https://user-images.githubusercontent.com/9111900/141754649-8f809e25-fde0-463d-9fcc-9a9d658e3e8b.png)

BrOstro:
Getting the same issue as OP on Windows 10

ausmreed:
For anyone having this issue, I followed these steps from another issue and was able to log in: 

1. Open up YTMD
2. Click "Sign in", nothing
3. Right-click system tray icon and select Guest
4. Second window opens up with login
5. Sign in via second window, still guest in main window
6. Close YTMD fully and reopen
[...]

PR #1108, opened 2023-04-01
Remove user agent, allows Google Sign In with 2FA

joser93:
Greetings

I just saw this project on a video and find out about this issue https://github.com/ytmdesktop/ytmdesktop/issues/866. 

As far as I see, the user agent is something that you want to keep, but I don't see why, so, I tested to remove it at all, Also I tested with 1.13 but I am not submitting a PR of that just to avoid merge conflicts in the future or maybe a 1.13.1 branch and release for this fix.

Anyway, I tested this, used npm start and I am able to log in, the 2FA will ask for a hardware key, but at the bottom of that, there's another option to choose your desired verification process, choose what you can use and sign in. 

I also saw on the issue that changing this might causes some troubles, but, with 1.13, I was not able to spot a problem, playback, shortcuts, lyrics, sysicon tray, all seems to works fine. Although, testing 1.14 (master) I encounter other issues that are not related with this change, maybe I'll post about it later. If there's something else that I must test or contribute something else to fix this, I'll keep in touch.

I hope this helps and brings back on track YTM development.

### DIFF
--- a/main.js
+++ b/main.js
@@ -273,9 +273,7 @@ async function createWindow() {
             const newRequestHeaders = Object.assign(
                 {},
                 details.requestHeaders || {},
-                {
-                    'User-Agent': settingsProvider.get('user-agent'),
-                }
+                {}
             )
             callback({ requestHeaders: newRequestHeaders })
         }
@@ -1432,8 +1430,7 @@ async function createWindow() {
                 './src/pages/shared/window-buttons/window-buttons.html'
             ),
             {
-                search:
-                    'page=settings/sub/last-fm/last-fm-login&icon=music_note&hide=btn-minimize,btn-maximize&title=Last.FM Login',
+                search: 'page=settings/sub/last-fm/last-fm-login&icon=music_note&hide=btn-minimize,btn-maximize&title=Last.FM Login',
             }
         )
     }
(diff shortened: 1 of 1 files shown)


Alipoodle:
This feels like a step back 😕
The reason for this feature allows for you to customise it if you wanted to, but even with setting

If anything you can just Overwrite the existing Useragent that WE provide with one and this should work.
However I suspect that this isn't a "perfect" fix and will only work temporarily.

joser93:
Thanks for the reply, but I didn't get the reason, the idea is to customize the user agent because it's allowed to do that, allows another feature to work, gives long term support without building a new version when is needed to change it or something else?

Anyway, i get that this is a feature that some people are interested about, but, should be a default user agent on first start or can be an option to allow in the settings that you can toggle and change it? the goal here is the main issue is solve, allow sign in with 2FA users and fixes a bug that some users might not use the app because of this. 

A decision should be made in the greater good for everyone or the majority and keep the development of the app, in my humble opinion.

If there's any feature that I must test without user agent, please, let me know.

fimion:
As a side note: using 1.14.2 on Macos 13.5 and setting the user-agent to `null` in the config.json file does seem to  fix this issue.

Another thing i noticed too is that the back arrow on the login page was centered vertically when the login worked, but was always in the upper left hand corner when the login didn't work.

Alipoodle:

Issue #1106, opened 2023-03-28
non-working button in the bottom panel (+Add to playlist)


Mee8pon:
### Consent

- [X] I verified that there is no open/closed issue for the same subject.
- [X] I understand that **YTMDesktop have NO affiliation with Google or YouTube**

### Current Behavior

Click - Not working

### Expected Behavior

Should open the panel for adding music to a playlist

### Steps To Reproduce

1. Turn on any music
2. Pay attention to the bottom panel (where there is play, stop, next and previous music)
3. Click on the "add to playlist" button (not in the "more" panel)
It won't work

### YTMDesktop

1.3.0

### OS

Windows

### OS Version

10

### Arch

x64

### Installation way

.exe

### Anything else?

![image](https://user-images.githubusercontent.com/129200349/228290008-8ca72f54-6f47-4ca9-84da-57d21744440d.png)
It is worth noting that in the web version there is no such button at all

![image](https://user-images.githubusercontent.com/129200349/228290339-b35ba630-4ab9-4d01-8532-45ed1456ca37.png)

Hershey025:
Even the add to library option doesn't work. It just doesn't show the option even after enabling it from the settings.

Alipoodle:
This is due to the icon change. A fix is needed to how we should fetch this.

Alipoodle:
Closing issue due to no longer supporting YouTube Music Desktop v1 + Release of YouTube Music v2.
Please re-open a NEW issue if this is still occurring in the latest build of the application, however as a re-write has been done on the entire project most issues will be fixed, no longer a feature or re-worked.

PR #1117, opened 2023-05-22
:bug: fix fetching playlists from popup

ce-dr:
Looks like they changed the icon used in the popup menu. I replaced all mentions of the old one. This should solve #1106  and potentially other issues regarding the infoplayer as well


Alipoodle:

PR #1155, opened 2023-09-07
Update general settings labels in en.json

CodeManMike:
- Update "Discord Rich Presence" label to "Enable Rich Discord Presence"
- Update "Keep YTMDesktop on background when window close" label to "Keep YTMDesktop in the background when window is closed"


Alipoodle:
Strings are no longer used in this Repo (and have come back somehow)

Please see https://github.com/ytmdesktop/ytmdesktop-locales instead for making a PR to change locales as this is the one which is dynamically taken inside of v1.14 versions of the app.

_While my knowledge of v1.14 is kinda old now, given I've been working on v2 for a Long while now_
`Discord Rich Presence` -> `Enable Discord Rich Presence` seems like an unneeded change 😅 
The settings (And even integrations section) is mostly considered to already be understood of enabling and disabling features with the Switches on the right hand side, and most things don't prefix themselves with Enable. Some things prefix themselves with Disable such as `Disable Hardware encoding` as this is an Inverted statement to the switch.

`Keep YTMDesktop in the background when window is closed` Otherwise does sound like english.

PR #1169, opened 2023-09-23
👷 Better build parallelization

ElectricalBoy:
This PR adds:

- Further parallelism in the build workflow
- Enables (cross-)compilation for arm64 (possibly resolves #572)
  I own a Surface Pro X (a niche platform IMO... lol) myself and can confirm that a "working" build for `windows/arm64` is generated with the changed workflow. See https://github.com/ElectricalBoy/ytmdesktop/actions/runs/6281142179 for POC.

PR #1182, opened 2023-10-18
📝 Add ARM64 to Bug Report Template

ElectricalBoy:
Now that ARM64 builds are available since [v2.0.0-rc.4](../releases/tag/v2.0.0-rc.4) (via #1169), it seems appropriate to add ARM64 to the architecture options in the bug report form. Thus, this PR adds ARM64 option to the bug report form.


Alipoodle:
😋 Is it possible to migrate this to the Development branch now?

PR #1228, opened 2023-12-30
✨🐛Auto click "Skip button" bug fixes for adPlaying state

DamienStaebler:
- ⭐ Auto ad skipper ( When skip button shows )
- Fixed `adPlaying` state never change back to false.


src/main/index.ts
```diff
@@ -624,6 +625,20 @@ function setupTaskbarFeatures() {
         mode: isPlaying ? "normal" : "paused"
       });
     }
+
+    if (store.get("playback.autoClickSkipAd") && state.adPlaying) {
```
Alipoodle, line 629: This section is contained within the Setting up of Taskbar features 😅
I feel like this should be placed in a better location

DamienStaebler: solved👍

banocean:
I think there would be an easier way to achieve that and implement ad-blocking at the same time, using an already good ad-block (like uBlock Origin) and [loading it as a Chrome extension](https://www.electronjs.org/docs/latest/api/extensions). Changing way to handle it seems as a good way to develop further app functionality that would avoid dealing with problems caused by frontend changes

DamienStaebler:
Blocking ads completely might put this project on Google's watch list. My main purpose of creating this PR is to support those authors but also able to enjoy the music when you are doing another task unable to skip a 5 minute ads that are playing.

DamienStaebler:
Youtube Music made it impossible to simulate click due to isTrusted. I don't see a clear way to bypass it so I am gonna close this PR

PR #1250, opened 2024-01-29
🐛  Fix undefined ReferenceError: open is not defined

BioCla:
As I was getting the below error:

```bash
ReferenceError: open is not defined
    at App.<anonymous> (/path/to/ytmd/.webpack/main/index.js:65175:31)
```

after analyzing the error I explicitly added the fs method call, which solved the problem.


NovusTheory:
This bug was already fixed in the development branch. Development branch is also the home to v2 now as it has made its official release

v2 branch may get deleted in the near future as a result of that change

PR #1267, opened 2024-01-31
✨ Add basic support for chrome extensions

bennett-sh:
This PR adds an extension folder (```[app directory]/extensions```). You can paste unpacked extensions into it and the app will load them. Note that this is quite a barebones implementation as there isn't any UI yet so the user has to manually find the folder.

Also, the functionality in extensions only goes as far as electron supports it. You can read about which features aren't working yet [here](https://www.electronjs.org/docs/latest/api/extensions).

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1816,8 +1816,8 @@ app.on("ready", async () => {
   }
 
   // Load Chrome Extensions
-  log.info("Load chrome extensions");
-  loadChromeExtensions();
+  log.info("Loading chrome extensions");
+  await loadChromeExtensions();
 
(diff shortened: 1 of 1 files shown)


Alipoodle:
Thank you for your PR

We appreciate this PR for adding support for Chrome Extensions, however in the current state of it being a blanket open for people to put in any Extension and with no User Experience in regards to Interacting, Installing and more we do not feel like we can pull this PR.

We would like to have a bit more control over making sure that user safety is still possible to maintain and that a user doesn't make a "YTMD Adblocker" which is just grabbing people's Google accounts and sending to some suspious server.

PR #1283, opened 2024-02-05
🐛 Fix: copy and paste on MacOS

### LINKED ISSUE #1276: Can't paste email or password - Electron Menu issue
### Consent

### Current Behavior

Trying to paste my email or password into Google login page does nothing

### Expected Behavior

I should be able to paste my email and password into the login screen

### Steps To Reproduce

1. Download and open
2. Hit sign in

### YTMDesktop

v2.0.0

### OS

MacOS

### OS Version

14.4

### Arch

ARM (Apple silicon)

### Installation way

dropped into Applications folder

### Anything else?

Double annoying because of https://github.com/ytmdesktop/ytmdesktop/issues/1227 where I have to log in each time I close the app.

The form for this bug submission only has dropdowns for `x64` and `x86` for the Arch section

jaredkotoff:
Fixes #1276 
Reference: https://github.com/onmyway133/blog/issues/67#issuecomment-459846635

Screenshot 2024-02-06 at 3 00 40 PM

Doesn't fix support for 1password autofill though

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -47,6 +47,7 @@ declare const YTMD_UPDATE_FEED_OWNER: string;
 declare const YTMD_UPDATE_FEED_REPOSITORY: string;
 
 const assetFolder = path.join(process.env.NODE_ENV === "development" ? path.join(app.getAppPath(), "src/assets") : process.resourcesPath);
+const isDarwin = process.platform === "darwin";
 
 let applicationExited = false;
 let applicationQuitting = false;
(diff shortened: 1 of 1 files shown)


src/main/index.ts
```diff
@@ -128,9 +129,67 @@ log.info("Application launched");
 // Enforce sandbox on all renderers
 app.enableSandbox();
 
-// Disabling the application menu improves performance, but we're disabling it to patch out default hotkeys for now
-// TODO: Come back to this at a future point in time to determine the necessary need for an application menu
```
jaredkotoff, line 132: Not sure about this though

Alipoodle: https://www.electronjs.org/docs/latest/tutorial/performance#8-call-menusetapplicationmenunull-when-you-do-not-need-a-default-menu 🤷

https://github.com/electron/electron/issues/35512

src/main/index.ts
```diff
@@ -128,9 +129,67 @@ log.info("Application launched");
...
+    },
+    {
+      label: "Quit",
+      accelerator: "Command+Q",
+      click: () => {
+        app.quit();
+      }
+    }
+  ]
+};
```
NovusTheory, line 150: Could this just be done with just `{ role: "appMenu" }`? I'm not sure what this provides on mac and whether it may be more useful or have too much that the application doesn't need or some functionality of it end up broken.

Side note `appMenu` is ONLY available on macOS so that needs to be platform guarded.

src/main/index.ts
```diff
@@ -128,9 +129,67 @@ log.info("Application launched");
...
+      accelerator: "CmdOrCtrl+V",
+      role: "paste"
+    },
+    {
+      label: "Select All",
+      accelerator: "CmdOrCtrl+A",
+      role: "selectAll"
+    }
+  ]
+};
```
NovusTheory, line 189: Same as before with can this be done by just doing `{ role: "editMenu" }`?

src/main/index.ts
```diff
@@ -128,9 +129,67 @@ log.info("Application launched");
...
+
+const template = [application, edit];
+Menu.setApplicationMenu(Menu.buildFromTemplate(template));
```
NovusTheory, line 192: Since other operating systems don't need an application menu, and at least on windows, we don't even show one I would platform guard setting an application menu to mac and explicitly set it to null on other operating systems.

NovusTheory (changes requested):

src/main/index.ts
jaredkotoff: Done, the only thing is if you want to deviate from the defaults for hotkey purposes

NovusTheory (approved):
LGTM as long as it's doing what it should

Alipoodle (approved):

Issue #533, opened 2020-12-15
[Feature Request] Last.fm scrobble percentage setting


ObviouslyGreen:
- [x] I understand that **YTMDesktop have NO affiliation with Google or YouTube**.
- [x] I verified that there is no open issue for the same subject.

**Is your feature request related to a problem? Please describe.**
Sending scrobbles for last.fm for songs I skip

**Describe the solution you'd like**
Common for scrobblers to have a setting to only scrobble when playback has reached x% (usually configurable between 50-100% of playback)

**Describe alternatives you've considered**
Can maybe just make the threshold a little larger than what it is instead of making it configurable, since it currently just [scrobbles at 1%](https://github.com/ytmdesktop/ytmdesktop/blob/931d40d856e3d59745cbba62f8c3928ce8f09293/main.js#L400)

**Additional context**
Example of setting in other scrobblers
![image](https://user-images.githubusercontent.com/4269549/102266532-6fcf3380-3ecd-11eb-94ac-d70260f4bf8f.png)

Have been enjoying using the app, thanks!

eddielenton:
I have no idea how to do that, but I have submitted a pull request for changing the default value to 20% at least as a work around for now.
https://github.com/ytmdesktop/ytmdesktop/pull/561

PR #1286, opened 2024-02-06
✨🐛 Add Scobble percent & Fix Last.fm login

jaredkotoff:
* fixes login for lastfm
* add user selectable scrobble percentage (https://github.com/ytmdesktop/ytmdesktop/issues/533)

Screenshot 2024-02-08 at 10 11 21 AM


Alipoodle:
Only skimmed though the PR currently.

But why the change to 5 minutes when the documentation for Last.fm specifically notes 4 minutes:

> "And the track has been played for at least half its duration, or for 4 minutes (whichever occurs earlier.)"
[Last.fm Document](https://www.last.fm/api/scrobbling#:~:text=And%20the%20track%20has%20been%20played%20for%20at%20least%20half%20its%20duration%2C%20or%20for%204%20minutes%20(whichever%20occurs%20earlier.))

src/main/integrations/last-fm/index.ts
```diff
@@ -97,14 +114,14 @@ export default class LastFM implements IIntegration {
...
+      const scrobblePercentDecimal = this.lastfmDetails.scrobblePercent / 100;
       const scrobbleTimeRequired = Math.min(
-        // Scrobble the track if it has been played for more than 50% of its duration
-        Math.round(state.videoDetails.durationSeconds / 2),
-        // OR if it has been played for more than 4 minutes
-        240
+        // Scrobble the track if it has been played to the percent picked by the user
+        Math.round(state.videoDetails.durationSeconds * scrobblePercentDecimal),
+        // OR if it has been played for more than 5 minutes
+        300
```
Alipoodle, line 122: Why has this changed to 5 minutes?

As per Last.fm specs:
> "And the track has been played for at least half its duration, or for 4 minutes (whichever occurs earlier.)"
[Last.fm Document](https://www.last.fm/api/scrobbling#:~:text=And%20the%20track%20has%20been%20played%20for%20at%20least%20half%20its%20duration%2C%20or%20for%204%20minutes%20(whichever%20occurs%20earlier.))

Alipoodle, line 122: Why has this changed to 5 minutes?

As per Last.fm specs:
> "And the track has been played for at least half its duration, or for 4 minutes (whichever occurs earlier.)"
[Last.fm Document](https://www.last.fm/api/scrobbling#:~:text=And%20the%20track%20has%20been%20played%20for%20at%20least%20half%20its%20duration%2C%20or%20for%204%20minutes%20(whichever%20occurs%20earlier.))

src/main/integrations/last-fm/index.ts
```diff
@@ -40,12 +40,30 @@ export default class LastFM implements IIntegration {
...
+
+      authWindow.on("closed", function () {
+        authWindow = null;
+        resolve(true);
+      });
+    });
   }
 
   private async getSession() {
+    await this.authenticateUser();
```
Alipoodle, line 66: This appears to be requiring a user to get a new Token on each opening of YTM Desktop and not using their existing token from a previous session 🤔

The reason for **not** doing it here, but instead doing it in the `json.error` is to know when the token is invalid and then get a new valid token to use for that session.

src/main/integrations/last-fm/index.ts
```diff
@@ -40,12 +40,30 @@ export default class LastFM implements IIntegration {
     this.lastfmDetails.token = await this.createToken();
     this.saveSettings();
 
-    shell.openExternal(
-      `https://www.last.fm/api/auth/` + `?api_key=${encodeURIComponent(this.lastfmDetails.api_key)}` + `&token=${encodeURIComponent(this.lastfmDetails.token)}`
-    );
+    let authWindow = new BrowserWindow({
```
Alipoodle, line 43: Curious to the inclusion of using a Browser window and not using the user's own Browser which will have their Login/Password already stored and ready.
I feel that from both a User perspective this also seems like a better "secure" option considering it's not being processed through YTMD.

Alipoodle (commented):
Leaving this as comments for now until I can test and verify the Authenicate thing.

The Ability to have the progress of the feature looks perfect and it's a lot simplier than I originally expected it to be, but some of the changes break L.fm's documentation / seem a step back.

src/main/integrations/last-fm/index.ts
jaredkotoff: Ah i didnt realize this was a magic number, ill rever it

src/main/integrations/last-fm/index.ts
jaredkotoff: The reason for this was to track when the user closed the browser after allowing access for step `2.3. Request authorization from the user` which must be done before we fetch a user session for scrobbling.  
Before this change the token was returning an error as it wasn't authorized to a user.
I admit i did this pretty hastily, but it works. I'm open to input but the current code is simply broken and does things out of order.

src/renderer/windows/settings/Settings.vue
```diff
@@ -400,6 +403,18 @@ window.ytmd.handleUpdateDownloaded(() => {
               </p>
             </div>
           </div>
+          <div v-if="lastFMEnabled" class="setting indented">
+            <YTMDSetting
+              v-model="scrobblePercent"
+              type="range"
+              name="Scrobble percent: "
+              description="Determines when a song is scrobbled"
+              min="50"
```
jaredkotoff, line 412: Also setting this to be the minimum per the guidelines
> track has been played for at **least** half its duration

src/main/integrations/last-fm/index.ts
jaredkotoff: Ok digging into this. These changes might not be necessary, it seems that no matter what that the token in the store is never getting updated, which is the core of my issues since it just keeps trying to use the same token over and over. 

Trying to figure out what's causing this, and then will re-evaluate changes to the flow

src/main/integrations/last-fm/index.ts
jaredkotoff: Ok I was able to revert this change once I found the **real** bug, this is still a little janky and I think requires the user to toggle the setting off and then back:

If the token fails: get a new token and reauth them, but a new session isn't created until the integration switch is set off and then enabled again

src/main/integrations/last-fm/index.ts
```diff
@@ -295,9 +300,13 @@ export default class LastFM implements IIntegration {
 
   private async saveSettings(): Promise<void> {
     try {
-      this.store.set("lastfm.sessionKey", safeStorage.encryptString(this.lastfmDetails.sessionKey).toString("hex"));
-      this.store.set("lastfm.token", safeStorage.encryptString(this.lastfmDetails.token).toString("hex"));
-    } catch {
+      if (this.lastfmDetails.sessionKey) {
+        this.store.set("lastfm.sessionKey", safeStorage.encryptString(this.lastfmDetails.sessionKey).toString("hex"));
+      }
```
jaredkotoff, line 305: This was the source of the problem. `sessionKey` was null and throwing a type error and not updating the token in the store. 
We don't have a session key, so we can't get a session key => infinite loop

src/renderer/windows/settings/Settings.vue
```diff
@@ -229,6 +233,13 @@ function checkForUpdates() {
   checkingForUpdate.value = true;
 }
 
+async function logoutLastFM() {
+  store.set("lastfm.sessionKey", null);
+  lastFMEnabled.value = false;
+  lastFMSessionKey.value = null;
+  await settingsChanged();
+}
```
jaredkotoff, line 241: Adding this for debugging but also seems like a nice QoL.
The only thing I couldn't get working nicely is the `User is Authenticated:` below to update automatically when the user auth's

src/renderer/windows/settings/Settings.vue
```diff
@@ -396,10 +407,23 @@ window.ytmd.handleUpdateDownloaded(() => {
               <p class="description">
                 User is Authenticated:
                 <span v-if="lastFMSessionKey" style="color: #4caf50">Yes</span>
+                <button v-if="lastFMSessionKey" class="lastfm-logout-button" @click="logoutLastFM">Logout</button>
```
Alipoodle, line 410: image

If possible could you have this to look like the Discord rich presences connection failed.
```html
          <div v-if="discordPresenceEnabled && discordPresenceConnectionFailed" class="setting indented">
            <p class="discord-failure">Discord connection could not be established after 30 attempts</p>
            <button @click="restartDiscordPresence">Retry</button>
          </div>
```
I suspect you just need to put the button outside of the `<p>` tag

src/renderer/windows/settings/Settings.vue
```diff
@@ -396,10 +407,23 @@ window.ytmd.handleUpdateDownloaded(() => {
...
                 <span v-else style="color: #ff1100">No</span>
               </p>
             </div>
           </div>
+          <div v-if="lastFMEnabled" class="setting indented">
+            <YTMDSetting
```
Alipoodle, line 416: Is it possible to get the range to be right aligned
image

Similar to the zoom. I suspect that this is a problem with the range element and the intended class currently.
image

src/renderer/windows/settings/Settings.vue
```diff
@@ -396,10 +407,23 @@ window.ytmd.handleUpdateDownloaded(() => {
...
+              v-model="scrobblePercent"
+              type="range"
+              name="Scrobble percent: "
+              description="Determines when a song is scrobbled"
+              min="50"
+              max="100"
```
Alipoodle, line 422: Worth checking that it works correctly at 100%.
As obviously the timer might last a little bit too long before it's then cancelled from the next song playing.

src/main/integrations/last-fm/index.ts
Alipoodle: Still need to look into this, but this looks more sound than a try + catch. (this though that's still worth including)

Alipoodle (changes requested):
Primarily changes for the settings to make the UI look more consistent with the rest of stuff.

I will get this downloaded + tested tomorrow to see about the session stuff and the feature works as expected.
**But otherwise this is looking good.**

src/renderer/windows/settings/Settings.vue
jaredkotoff: I couldn't figure out how to get devtools on this settings windows look debug any of this. Any ideas?

src/renderer/windows/settings/Settings.vue
jaredkotoff: What do you think about limiting this to `95` or subtracting a second from the timer: ` Math.round(state.videoDetails.durationSeconds * scrobblePercentDecimal) -1`

src/main/integrations/last-fm/index.ts
jaredkotoff: the try catch is still there, just does a check so one erroring doesnt break the other from updating

src/main/integrations/last-fm/index.ts
Alipoodle: Session is grabbed on next time a song is played, so it's not entirely ignored and a problem.
I do want to add a `refresh` button to the Settings page, but I need to do some exposing of the APIs from the integration to the settings.

src/main/integrations/last-fm/index.ts
Alipoodle: I saw the Try + Catch is still there but checking that this isn't breaking anything. But it's all good

src/renderer/windows/settings/Settings.vue
Alipoodle: I did find a bug. If the value for Last.FM progress is changed during a session, as the L.fm settings are cached in the extension, the progress value isn't updated until a Integration is turned off/on OR Program is restarted.

This would be ideal to update this value.

src/main/index.ts
```diff
@@ -361,7 +361,8 @@ const store = new Conf<StoreSchema>({
       api_key: "2a69bcf769a7a28a8bf2f6a5100accad",
       secret: "46eea23770a459a49eb4d26cbf46b41c",
       token: null,
-      sessionKey: null
+      sessionKey: null,
+      scrobblePercent: 50
```
Alipoodle, line 365: @NovusTheory can probably give a better answer on this. But while testing this, was given an error for the `delete` when no value is set.

 I suspect that a Migration will need to be added for this value for it to apply the default value for people who are using YTMD.

NovusTheory: @Alipoodle You are correct that this needs a migration but it ideally would go under the next version that this would get merged into. Us maintainers can mark a note and just add it before the next version comes but this seems problematic long term.

Since I'm at least currently planning for 2.0.1 to correct all the bugs that have been introduced this probably can slip under that with a minor feature addition which happens to have bug fixes.

Alipoodle: We've discussed about what is the best solution in regards to doing migration and for now we're happy for user's to include it in the PR 😄 

We might change this up in the future if things don't make it into the next release etc.

Alipoodle (changes requested):
Final change for the `.vscode/settings.json` file. And otherwise this is looking good, and should make it for the v2.0.1 release.

Alipoodle (approved):
All looking good :)

PR #1294, opened 2024-02-08
✨ListenBrainz intergration

ZedAU:
This pull request adds scrobble support for ListenBrainz

In the current state it works, But I require help or ideas on two things. getting the token hidden in the settings when one is present, and help with styling to fit with the apps other elements in the settings menu


Sancoraa:
I have a suggestion to make about this feature.

We should be able to change the default apiRoot for users that uses self-hosted music scrobble database.
I for example use [Maloja](github.com/krateng/maloja) which supports the ListenBrainz API.
I had to hard change the API endpoint myself in the ytmdesktop config file.

Either that, or best scenario, implement the Maloja API into the application.

PR #1355, opened 2024-05-08
✨ Add change song command to API

Setadokalo:
This just adds a command to the API to change the current song/playlist, using logic copied from the existing implementation for the URI handler. There's still useful functionality that won't be covered, such as adding a song to the queue without replacing the queue entirely, but this at least covers a major use case.

There's no validation to ensure the video and playlist id are correct, which can cause some weirdness; invalid playlist ids will cause the bottom bar to disappear, for example.

### DIFF
--- a/src/main/integrations/companion-server/api-shared/errors.ts
+++ b/src/main/integrations/companion-server/api-shared/errors.ts
@@ -6,6 +6,7 @@ const errorCodes = [
   "INVALID_REPEAT_MODE",
   "INVALID_SEEK_POSITION",
   "INVALID_QUEUE_INDEX",
+  "INVALID_CHANGE_REQUEST",
   "UNAUTHENTICATED",
   "AUTHORIZATION_DISABLED",
   "AUTHORIZATION_INVALID",
@@ -21,6 +22,7 @@ export const InvalidVolumeError = createError<[number]>("INVALID_VOLUME", "Volum
 export const InvalidRepeatModeError = createError<[RepeatMode]>("INVALID_REPEAT_MODE", "Repeat mode '%s' cannot be set", 400);
 export const InvalidPositionError = createError<[number]>("INVALID_SEEK_POSITION", "Seek position '%s' is invalid", 400);
 export const InvalidQueueIndexError = createError<[number]>("INVALID_QUEUE_INDEX", "'%s' is an invalid queue index for the current queue", 400);
+export const InvalidChangeVideoRequestError = createError<[]>("INVALID_CHANGE_REQUEST", "'videoId', 'playlistId', or both must be provided", 400);
 export const UnauthenticatedError = createError<[]>("UNAUTHENTICATED", "Authentication not provided or invalid", 401);
 export const AuthorizationDisabledError = createError<[]>("AUTHORIZATION_DISABLED", "Authorization requests are disabled", 403);
 export const AuthorizationInvalidError = createError<[]>("AUTHORIZATION_INVALID", "Authorization invalid", 400);
--- a/src/main/integrations/companion-server/api-shared/schemas.ts
+++ b/src/main/integrations/companion-server/api-shared/schemas.ts
@@ -56,6 +56,13 @@ export const APIV1CommandRequestBody = Type.Union([
     })
   }),
   Type.Object({
+    command: Type.Literal("changeVideo"),
+    data: Type.Object({
+      videoId: Type.Optional(Type.String()),
+      playlistId: Type.Optional(Type.String())
+    })
+  }),
+  Type.Object({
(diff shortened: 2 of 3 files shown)


src/main/integrations/companion-server/api/v1/index.ts
```diff
@@ -148,6 +149,22 @@ const CompanionServerAPIv1: FastifyPluginCallback<CompanionServerAPIv1Options> =
           break;
         }
 
+        case "changeSong": {
```
NovusTheory: A better name such as `changeVideo` might be suited for this as we don't use the term `song` anywhere and neither does YTM

src/main/integrations/companion-server/api/v1/index.ts
```diff
@@ -148,6 +149,22 @@ const CompanionServerAPIv1: FastifyPluginCallback<CompanionServerAPIv1Options> =
...
+          const videoId = commandRequest.data.song;
+          const playlistId = commandRequest.data.playlist;
+          console.log("video: ", videoId, "playlist: ", playlistId);
```
NovusTheory: Remove this debug console log here

src/main/integrations/companion-server/api-shared/errors.ts
```diff
@@ -21,6 +21,7 @@ export const InvalidVolumeError = createError<[number]>("INVALID_VOLUME", "Volum
 export const InvalidRepeatModeError = createError<[RepeatMode]>("INVALID_REPEAT_MODE", "Repeat mode '%s' cannot be set", 400);
 export const InvalidPositionError = createError<[number]>("INVALID_SEEK_POSITION", "Seek position '%s' is invalid", 400);
 export const InvalidQueueIndexError = createError<[number]>("INVALID_QUEUE_INDEX", "'%s' is an invalid queue index for the current queue", 400);
+export const InvalidSongRequestError = createError<[]>("INVALID_SONG_REQUEST", "Song and Playlist cannot both be empty", 400);
```
NovusTheory: `InvalidChangeVideoRequestError` to align with requested change.

Recommended error details should be `"'id', 'playlistId', or both must be provided"`

src/main/integrations/companion-server/api-shared/schemas.ts
```diff
@@ -55,6 +55,13 @@ export const APIV1CommandRequestBody = Type.Union([
       minimum: 0
     })
   }),
+  Type.Object({
+    command: Type.Literal("changeSong"),
```
NovusTheory: `changeVideo` for requested change

src/main/integrations/companion-server/api-shared/schemas.ts
```diff
@@ -55,6 +55,13 @@ export const APIV1CommandRequestBody = Type.Union([
...
+    data: Type.Object({
+      song: Type.Optional(Type.String()),
```
NovusTheory: `id` for requested change

src/main/integrations/companion-server/api-shared/schemas.ts
```diff
@@ -55,6 +55,13 @@ export const APIV1CommandRequestBody = Type.Union([
...
+      playlist: Type.Optional(Type.String())
```
NovusTheory: `playlistId` for readability and consistency purposes

NovusTheory (changes requested):
Just requesting that the command name be changed and it should be good.

Also please make sure to add the error code to the `errorCodes` array that was above otherwise the API will not properly return it https://github.com/ytmdesktop/ytmdesktop/pull/1355/files#diff-84b5d03309d58f0086ac6d2cb4cdce5f610a13d9f697edfb31974dd3a8c2c9b2R4-R17

src/main/integrations/companion-server/api/v1/index.ts
Setadokalo: Done.

src/main/integrations/companion-server/api/v1/index.ts
Setadokalo: removed

Setadokalo:
Done, with the slight exception of changing `song` to `id`; given that either the playlist id or video id can be provided, it feels weird to me to give video id the 'default' name, so I changed it to `videoId` and `playlistId`. It looks like that's consistent with internal stuff as well.

NovusTheory (approved):
LGTM

Setadokalo:
Squashed the changes into a single commit to make it ready to merge.

Alipoodle (approved):
Shocked as to how simple this is, but looks and feels effective :)
Interesting to see how people would integrate with this.

PR #1398, opened 2024-07-23
Refactor and Fix Log Hook in 'index.ts'

Akiisqt:
This is a small fix/improvement to the existing Log Hook in 'index.ts'

The original code errored if message.data was not an array or contained non-string elements. The changes check the types to prevent errors from occurring.


Alipoodle:
Please can you fix the Title and include a description which correctly addresses what this PR does.

"Fix Index.ts" doesn't mean anything from a Glance.

I'd recommend looking at other PRs as examples of what to do.

Akiisqt:
Sorry, hopefully that's better now!

PR #1400, opened 2024-07-23
🩹Refactor and Fix Log Hook in 'index.ts'

### LINKED ISSUE #1394: Message data can be null, or the data inside can not be a string
### Consent

### Current Behavior

Electron is causing errors with the error handler

### Expected Behavior

error handler handles errors correctly

### Steps To Reproduce

I just used it

### YTMDesktop

Development

### OS

Windows

### OS Version

11

### Arch

x64

### Installation way

.exe

### Anything else?

Fix is to just add type checking:
main -> index.js 

```ts
log.hooks.push((message, transport) => {
  if (transport !== log.transports.file) return message;

  const isSpamMessage = (data: any) => {
    return typeof data === 'string' && data.includes("Third-party cookie will be blocked.");
  };

  if (message?.data?.[0] && isSpamMessage(message.data[0])) {
    return false;
  }

  if (Array.isArray(message.data)) {
    message.data = message.data.map(data => {
      if (typeof data === 'string') {
        return data.replaceAll(/(?<=((https|http):\/\/)?.{1,64}(\..{1,64})?\..{1,64}\/)([\S]+)/gm, "[REDACTED]");
      }
      return data;
    });
  }

  return message;
});
```

Akiisqt:
This is a small fix/improvement to the existing Log Hook in 'index.ts'

The original code errored if message.data was not an array or contained non-string elements. The changes check the types to prevent errors from occurring.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -64,12 +64,12 @@ log.transports.console.format = "[{processType}][{level}]{text}";
 log.transports.file.format = "[{y}-{m}-{d} {h}:{i}:{s}.{ms}][{processType}][{level}]{text}";
 log.eventLogger.format = "Electron event {eventSource}#{eventName} observed";
 
-const containsSpam = (data: unknown): boolean => typeof data === 'string' && /third-party cookie will be blocked\./i.test(data);
+const isSpamLogMessage = (data: unknown): boolean => typeof data === 'string' && /third-party cookie will be blocked\./i.test(data);
 log.hooks.push((message, transport) => {
   // If the transport is not a file transport then return as is
   if (transport !== log.transports.file) { return message; }
   // If there isnt message data, or the data isnt a string, or the data is spam from Youtube Music, return false
-  if (message?.data?.[0] && containsSpam(message.data[0])) return false;
+  if (message?.data?.[0] && isSpamLogMessage(message.data[0])) return false;
 
   // Check it is an array, then redact sensitive info
   message.data = message.data.map(data => {
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -65,14 +65,23 @@ log.transports.file.format = "[{y}-{m}-{d} {h}:{i}:{s}.{ms}][{processType}][{lev
 log.eventLogger.format = "Electron event {eventSource}#{eventName} observed";
 
 log.hooks.push((message, transport) => {
-  if (transport !== log.transports.file) return message;
+  if (transport !== log.transports.file) { return message; }
 
-  // Annoyingly spammed log from YTM that doesn't need to get added to the log file
-  if (message.data[0].includes("Third-party cookie will be blocked.")) return false;
-  for (let i = 0; i < message.data.length; i++) {
-    let dataString: string = message.data[i] as string;
-    dataString = dataString.replaceAll(/(?<=((https|http):\/\/)?.{1,64}(\..{1,64})?\..{1,64}\/)([\S]+)/gm, "[REDACTED]");
-    message.data[i] = dataString;
+  const isSpamMessage = (data: any) => {
+    return typeof data === 'string' && data.includes("Third-party cookie will be blocked.");
+  };
+
+  if (message?.data?.[0] && isSpamMessage(message.data[0])) {
+    return false;
+  }
+
+  if (Array.isArray(message.data)) {
+    message.data = message.data.map(data => {
+      if (typeof data === 'string') {
+        return data.replaceAll(/(?<=((https|http):\/\/)?.{1,64}(\..{1,64})?\..{1,64}\/)([\S]+)/gm, "[REDACTED]");
+      }
+      return data;
+    });
   }
 
   return message;
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -68,7 +68,7 @@ log.hooks.push((message, transport) => {
   // If the transport is not a file transport then return as is
   if (transport !== log.transports.file) { return message; }
   // If there isnt message data, or the data isnt a string, or the data is spam from Youtube Music, return false
-  if (message?.data?.[0] && typeof message.data[0] === 'string' && message.data[0].includes("Third-party cookie will be blocked.")) return false;
+  if (message?.data?.[0] && (typeof message.data[0] === 'string' && message.data[0].includes("Third-party cookie will be blocked."))) return false;
 
   // Check it is an array, then redact sensitive info
   if (Array.isArray(message.data)) {
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -66,11 +66,11 @@ log.eventLogger.format = "Electron event {eventSource}#{eventName} observed";
 
 log.hooks.push((message, transport) => {
   // If the transport is not a file transport then return as is
-  if (transport !== log.transports.file) { return message; }
+  if (transport !== log.transports.file) return message;
   // If there isnt message data, or the data isnt a string, or the data is spam from Youtube Music, return false
   if (message?.data?.[0] && (typeof message.data[0] === 'string' && message.data[0].includes("Third-party cookie will be blocked."))) return false;
 
-  // Check it is an array, then redact sensitive info
+  // loop throu the data and redact any sensitive info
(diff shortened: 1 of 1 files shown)


Akiisqt:
didnt mean to delete this mb

src/main/index.ts
```diff
@@ -65,14 +65,23 @@ log.transports.file.format = "[{y}-{m}-{d} {h}:{i}:{s}.{ms}][{processType}][{lev
...
 log.hooks.push((message, transport) => {
-  if (transport !== log.transports.file) return message;
-
-  // Annoyingly spammed log from YTM that doesn't need to get added to the log file
-  if (message.data[0].includes("Third-party cookie will be blocked.")) return false;
-  for (let i = 0; i < message.data.length; i++) {
-    let dataString: string = message.data[i] as string;
-    dataString = dataString.replaceAll(/(?<=((https|http):\/\/)?.{1,64}(\..{1,64})?\..{1,64}\/)([\S]+)/gm, "[REDACTED]");
-    message.data[i] = dataString;
+  if (transport !== log.transports.file) { return message; }
```
Alipoodle, line 70: Not sure why you've re-added these brackets?
There's a few instances where I feel like brackets shouldn't be used, and this is a key one.

It's clear to see that there's only 1 thing running and it's escaping instantly.

src/main/index.ts
```diff
@@ -65,14 +65,23 @@ log.transports.file.format = "[{y}-{m}-{d} {h}:{i}:{s}.{ms}][{processType}][{lev
...
+
+  const isSpamMessage = (data: any) => {
```
Alipoodle: I feel like this could be improved function name to `isSpamLogMessage`
as Message could imply lots of different things?

src/main/index.ts
```diff
@@ -65,14 +65,23 @@ log.transports.file.format = "[{y}-{m}-{d} {h}:{i}:{s}.{ms}][{processType}][{lev
...
+    return typeof data === 'string' && data.includes("Third-party cookie will be blocked.");
+  };
+
+  if (message?.data?.[0] && isSpamMessage(message.data[0])) {
+    return false;
+  }
```
Alipoodle: I feel like the `isSpamMessage` function migth be better moved to a different place, and having the


```js
if (transport !== log.transports.file) return message;
if (message?.data?.[0] && isSpamMessage(message.data[0])) return false;
````
next to it to be the clear "Early escapes" from the function grouped together.

**Comment by Alipoodle** [comment]
I've currently yet to see an example of when the `message.data` has been a Non array element?
has there been an example for this you could provide?

Intellisense suggests that it will always be an array, It can be of any type variable, but must be an array of them
image

File: src/main/index.ts [hunk 4]
```diff
@@ -65,14 +65,23 @@ log.transports.file.format = "[{y}-{m}-{d} {h}:{i}:{s}.{ms}][{processType}][{lev
...
+
+  if (Array.isArray(message.data)) {
```
Alipoodle: As mentioned in my comment, this feels redundent given that this is always an array, and type checking for the message.data shouldn't be needed?

The contents of the array items inside of message.data should be tested they are given as `any`

Alipoodle (changes requested):
A few minor bits of clean up I feel to keep things consistent and better for general user readability.

src/main/index.ts
Akiisqt: Like I said, I had logged what it was sometimes it wasnt an array and it was spamming my console.. i can see if I can find the error

Akiisqt: `
[main][error] Electron event WebContents#did-fail-load observed {
  errorCode: -3,
  errorDescription: '',
  validatedURL: 'http://localhost:3000/main_window',
  isMainFrame: true,
  frameProcessId: 4,
  frameRoutingId: 1
}
Unhandled electron-log error TypeError: dataString.replaceAll is not a function
`

image


Once I added a check to see if it was a string, I got another error saying something wasn't an array and that just continued happening

src/main/index.ts
Akiisqt: That was just the name that you guys put in the comment

src/main/index.ts
Alipoodle: Are you able to debug and see when this happens to explain what is triggering this?

Otherwise I feel like something else is at fault here? Considering the `let dataString: string = message.data[i] as string;` should require it to become a non-nullable string and therefore provide the function to be available. 

I cannot see where this would actually trigger a problem 😖
And considering I've checked both my Personal + Laptop's Logs and yet to find this error to occur on either, I suspect this might be more a fault of your Fork at the moment.

Akiisqt:
`
Unhandled electron-log error TypeError: message.data[0].includes is not a function
    at C:\Users\Akiisqt\Downloads\temp\ytmdesktop\.webpack\main\index.js:49503:25
    at C:\Users\Akiisqt\Downloads\temp\ytmdesktop\.webpack\main\index.js:20701:24
    at Array.reduce (<anonymous>)
    at Logger.processMessage (C:\Users\Akiisqt\Downloads\temp\ytmdesktop\.webpack\main\index.js:20700:43)
    at Logger.logData (C:\Users\Akiisqt\Downloads\temp\ytmdesktop\.webpack\main\index.js:20661:10)
    at <computed> [as log] (C:\Users\Akiisqt\Downloads\temp\ytmdesktop\.webpack\main\index.js:20615:37)
    at C:\Users\Akiisqt\Downloads\temp\ytmdesktop\.webpack\main\index.js:49501:13
    at C:\Users\Akiisqt\Downloads\temp\ytmdesktop\.webpack\main\index.js:20701:24
    at Array.reduce (<anonymous>)
    at Logger.processMessage (C:\Users\Akiisqt\Downloads\temp\ytmdesktop\.webpack\main\index.js:20700:43)
[32184:0725/124746.069:ERROR:network_service_instance_impl.cc(600)] Network service crashed, restarting service.
`

src/main/index.ts
Alipoodle: This is still not addressed

src/main/index.ts
```diff
@@ -65,19 +65,23 @@ log.transports.file.format = "[{y}-{m}-{d} {h}:{i}:{s}.{ms}][{processType}][{lev
...
-  if (message.data[0].includes("Third-party cookie will be blocked.")) return false;
-  for (let i = 0; i < message.data.length; i++) {
-    let dataString: string = message.data[i] as string;
-    dataString = dataString.replaceAll(/(?<=((https|http):\/\/)?.{1,64}(\..{1,64})?\..{1,64}\/)([\S]+)/gm, "[REDACTED]");
-    message.data[i] = dataString;
-  }
+  // If the transport is not a file transport then return as is
+  if (transport !== log.transports.file) { return message; }
+  // If there isnt message data, or the data isnt a string, or the data is spam from Youtube Music, return false
+  if (message?.data?.[0] && (typeof message.data[0] === 'string' && message.data[0].includes("Third-party cookie will be blocked."))) return false;
```
Alipoodle: Why was this converted back to a line, I agreed that it looked nice for the function, especially as we have likely got other things to put into this list 😕

src/main/index.ts
```diff
@@ -65,19 +65,23 @@ log.transports.file.format = "[{y}-{m}-{d} {h}:{i}:{s}.{ms}][{processType}][{lev
...
+
+  // Check it is an array, then redact sensitive info
```
Alipoodle, line 74: Don't need to check if an array.

src/main/index.ts
Alipoodle: Just to say this is the only thing which still needs to be addressed, otherwise I feel like this is ready to merge.

Akiisqt:
Heya, were any of you able to comment on my last commit and see if thats what you were looking for?

Akiisqt:
This should be merged by now, I don't see a reason for it not to be.

Alipoodle:
> This should be merged by now, I don't see a reason for it not to be.

There's other things we're looking into.
However we're not pushing a new release just yet, as this is a bug fix it should be merged into the next one.

Please be patient as things aren't instantly moving along.

PR #1402, opened 2024-07-25
🩹 ✨ QoL changes to Discord presence integration

### LINKED ISSUE #1417: Discord rich presence displays song name twice where album name should be
### Consent

### Current Behavior

Discord Rich Presence shows:
Song Name
Artist Name
Song Name

Rather than:
Song Name
Artist Name
Album Name

example
example 2

### Expected Behavior

Album name should be displayed rather than displaying the song name twice 

### Steps To Reproduce

Go to Settings -> Integrations -> and enable Discord Rich Presence. Play literally any song and look at presence in Discord.

### YTMDesktop

2.0.5

### OS

Windows

### OS Version

11

### Arch

x64

### Installation way

.exe

### Anything else?

_No response_

Akiisqt:
Quick improvement to the current Discord RPC integration that fixes some issues I've noticed:

Trailing spaces
Simplified checks for when to update the activity
Moving setActivity into its own function to clean code
Renaming some functions to better suit what they do
removing buffering state
public variable to toggle listening or playing

### DIFF
--- a/src/renderer/windows/settings/Settings.vue
+++ b/src/renderer/windows/settings/Settings.vue
@@ -190,7 +190,7 @@ async function settingChangedFile(event: Event) {
     throw new Error("No setting specified in File Input");
   }
 
-  store.set(setting, target.files.length > 0 ? target.files[0].path : null);
+  store.set(setting, target.files.length > 0 ? window.ytmd.getTrueFilePath(target.files[0]) : null);
 
   target.value = null;
 }
--- a/src/renderer/@types/global.d.ts
+++ b/src/renderer/@types/global.d.ts
@@ -19,6 +19,7 @@ declare global {
       openSettingsWindow(): void;
       restartApplication(): void;
       restartApplicationForUpdate(): void;
+      getTrueFilePath(file: File): string;
 
       // Companion Authorization specific
       sendResult(authorized: boolean);
--- a/src/renderer/windows/settings/preload.ts
+++ b/src/renderer/windows/settings/preload.ts
@@ -1,7 +1,7 @@
 // See the Electron documentation for details on how to use preload scripts:
 // https://www.electronjs.org/docs/latest/tutorial/process-model#preload-scripts
 
-import { contextBridge, ipcRenderer } from "electron";
+import { contextBridge, ipcRenderer, webUtils } from "electron";
 import Store from "../../store-ipc/store";
 import { MemoryStoreSchema, StoreSchema } from "~shared/store/schema";
 import { WindowsEventArguments } from "~shared/types";
@@ -44,5 +44,6 @@ contextBridge.exposeInMainWorld("ytmd", {
   handleUpdateNotAvailable: (callback: (event: Electron.IpcRendererEvent) => void) => ipcRenderer.on("app:updateNotAvailable", callback),
   handleUpdateDownloaded: (callback: (event: Electron.IpcRendererEvent) => void) => ipcRenderer.on("app:updateDownloaded", callback),
   isAppUpdateAvailable: async (): Promise<boolean> => await ipcRenderer.invoke("app:isUpdateAvailable"),
-  isAppUpdateDownloaded: async (): Promise<boolean> => await ipcRenderer.invoke("app:isUpdateDownloaded")
+  isAppUpdateDownloaded: async (): Promise<boolean> => await ipcRenderer.invoke("app:isUpdateDownloaded"),
+  getTrueFilePath: (file: File) => webUtils.getPathForFile(file)
 });


Alipoodle:
> includes fixes from https://github.com/ytmdesktop/ytmdesktop/pull/1400

Please do not cross your PRs together 😖

src/main/integrations/discord-presence/index.ts
```diff
@@ -8,7 +8,18 @@ import { DiscordActivityType } from "./minimal-discord-client/types";
...
+  title: string;
+  author: string;
+  album?: string; // Optional
+  id: string;
+  trackState: VideoState;
+  thumbnail?: string; // Optional
+  progress: number;
+  duration: number;
+}
+// More appropriate name
```
Alipoodle: Please do not add comments in code for stuff like this.
You can specify that within the Description of the PR.

src/main/integrations/discord-presence/index.ts
```diff
@@ -21,7 +32,7 @@ function getHighestResThumbnail(thumbnails: Thumbnail[]) {
   }
   return url;
 }
-
+// Simplify these, when buffering it will just show as being paused
```
Alipoodle: Again comment is not needed.

src/main/integrations/discord-presence/index.ts
```diff
@@ -50,23 +53,15 @@ function getSmallImageText(state: number) {
...
 
-    case VideoState.Paused: {
-      return "Paused";
-    }
-
-    case VideoState.Buffering: {
-      return "Buffering";
-    }
-
     default: {
```
Alipoodle: We primarily have this as a bit more of a way of detecting when something is going wrong.
Especially if YouTube Music changes how these values work 😅

And providing as much information as possible doesn't seem unreasonable?

src/main/integrations/discord-presence/index.ts
```diff
@@ -8,7 +8,18 @@ import { DiscordActivityType } from "./minimal-discord-client/types";
 
 const DISCORD_CLIENT_ID = "1143202598460076053";
 
-function getHighestResThumbnail(thumbnails: Thumbnail[]) {
+interface VideoDetails {
```
Alipoodle: This is a direct 1-1 name of the existing Type called [`VideoDetails`](https://github.com/ytmdesktop/ytmdesktop/blob/development/src/main/player-state-store/index.ts#L30C1-L40C3).

Can this be renamed to reduce confusion between the two

src/renderer/windows/settings/Settings.vue
```diff
@@ -396,6 +399,14 @@ window.ytmd.handleUpdateDownloaded(() => {
...
             <button @click="restartDiscordPresence">Retry</button>
           </div>
+          <YTMDSetting 
+          v-model="discordPresenceListening" 
+          type="checkbox" name="Discord presence 
+          listening" @change="settingsChanged"/>
+          <div 
+          v-if="discordPresenceListening" 
+            class="setting indented">
+          </div>
```
Alipoodle: Missing Indentation
And please do not break inside of a Attribute.

I feel like that prettier was not run from this, and should run automatically before submitting a commit?

src/main/index.ts
```diff
@@ -65,15 +65,18 @@ log.transports.file.format = "[{y}-{m}-{d} {h}:{i}:{s}.{ms}][{processType}][{lev
...
+  // If there isnt message data, or the data isnt a string, or the data is spam from Youtube Music, return false
+  if (message?.data?.[0] && (typeof message.data[0] === 'string' && message.data[0].includes("Third-party cookie will be blocked."))) return false;
+
+  // Check it is an array, then redact sensitive info
+  message.data = message.data.map(data => {
+    if (typeof data === 'string') {
+      return data.replaceAll(/(?<=((https|http):\/\/)?.{1,64}(\..{1,64})?\..{1,64}\/)([\S]+)/gm, "[REDACTED]");
+    }
+    return data;
+  });
```
Alipoodle: As mentioned, please do not Cross your PRs.
1 PR = 1 Change of stuff.

You already have #1400 for this changes, and we will approve them seperately.

src/main/integrations/discord-presence/index.ts
```diff
@@ -84,86 +79,61 @@ export default class DiscordPresence implements IIntegration {
...
-      if (!this.lastTimeSeconds) this.lastTimeSeconds = timeSeconds;
-      if (!this.lastDuration) this.lastDuration = duration;
-      if (timeSeconds - this.lastTimeSeconds > 0) {
-        if (this.lastDuration === duration) {
-          // The time changed seconds but the duration hasn't. We use the old timestamp to prevent weird deviations
-          adjustedTimeSeconds = this.lastTimeSeconds;
-        } else {
-          this.lastDuration = duration;
-          this.lastTimeSeconds = timeSeconds;
+  public listeningActivityType : boolean = true; // change outside
```
Alipoodle: Please do not use this (and the setting of this in the `main/index.ts`)
And this also forces the listening presences upon restart even if the user has requested for the Playing.

Please instead provide this integration with the settings provider as is done with other ones, and add a listener for the settings change for this to update this variable.
Additionally probably worth doing a forced SetActivity as well, to update this for the user for better UX

Alipoodle (changes requested):
Some initial review of this before trying it out.

src/main/integrations/discord-presence/index.ts
Akiisqt: This is providing information to the whole world about your buffering issue, returning buffering as paused seems more reasonable.

src/main/index.ts
Akiisqt: Legit cannot console.log anything

Alipoodle: ?
As mentioned, don't mix your PRs. You already have submitted this code separately inside it #1400 . We don't need it to be committed twice.

src/renderer/windows/settings/Settings.vue
Akiisqt: Wdym, you want me to refactor the code or something? I believe I fixed it up a little bit tho

src/main/integrations/discord-presence/index.ts
NovusTheory: I actually do agree that since this is just presence for the public to know this isn't necessary information to provide. I know why we added it initially but I'm open for this to change.

src/main/integrations/discord-presence/index.ts
```diff
@@ -1,14 +1,16 @@
...
+import { MemoryStoreSchema, StoreSchema } from "~shared/store/schema";
+import { Unsubscribe } from "conf/dist/source/types";
 import DiscordClient from "./minimal-discord-client";
 import log from "electron-log";
 import { DiscordActivityType } from "./minimal-discord-client/types";
 
 const DISCORD_CLIENT_ID = "1143202598460076053";
 
-function getHighestResThumbnail(thumbnails: Thumbnail[]) {
+function getBestThumbnail(thumbnails: Thumbnail[]): string {
```
NovusTheory: I think the original function name is better suited as it's describing exactly what the function is doing.

src/main/integrations/discord-presence/index.ts
```diff
@@ -78,92 +64,72 @@ function stringLimit(str: string, limit: number, minimum: number) {
...
-              label: "Play on YouTube Music",
-              url: `https://music.youtube.com/watch?v=${state.videoDetails.id}`
-            },
-            {
-              label: "Play on YouTube Music Desktop",
-              url: `ytmd://play/${state.videoDetails.id}`
-            }
-          ]
-        });
+      if (oldState !== this.videoState) {
```
NovusTheory: This simplification shouldn't happen. The apps own player state store will never make a guarantee the state has to change for data to change.

Akiisqt: Have you tested it yet? This 'simplification' handles everything

NovusTheory: While it may work today it may not work tomorrow as we don't make nor want to make a guarantee the video state has to change for other data to change. This could cause the activity to display outdated data.

Akiisqt: The video always buffers when changing songs
If you want to be extra careful you can measure the change in progress on the video to see if it was moved fast or something to be extra careful

Alipoodle: I don't feel like you understand what he means.
This feature can work today with the current state of YouTube Music.

But we have no idea how YouTube Music's internals might update in the future, where progression between songs / tracks could be instant and never trigger this state to change. This is why we want more things to be considered for track changing.

src/renderer/windows/settings/Settings.vue
Alipoodle: While this is resolved, I want to explain.

We have [Prettier](https://prettier.io/) to automatically format the code to match closer to the style we want, and would address stuff like indentation and maximum line limit.
This also goes hand-in-hand with some of the other discussions I've had in DMs regarding what code editor you are using, and signed commits.

I suspect you are uploading the files directly to GitHub and avoiding the process which would automatically format the code.

src/main/integrations/discord-presence/index.ts
Akiisqt: I feel like alot more would break than just this

Alipoodle: https://github.com/ytmdesktop/ytmdesktop/blob/013ea5b4d6cecf35a49eaef30e92784057728b08/src/main/integrations/notifications/index.ts#L78-L79 - Notifications, Specifically checks that it's only playing, but then checks if the ID has changed.


https://github.com/ytmdesktop/ytmdesktop/blob/013ea5b4d6cecf35a49eaef30e92784057728b08/src/main/index.ts#L604 - Taskbar, Checking for if the track is playing for Progress on the Taskbar, other ones will show in pause state + Quick actions when hovering.


https://github.com/ytmdesktop/ytmdesktop/blob/013ea5b4d6cecf35a49eaef30e92784057728b08/src/main/integrations/last-fm/index.ts#L77-L80 - Last.FM, Checks if playing and from list of posibile IDs (Watching or Counterpart if using Premium) to make sure to not double scrobble.


None of these are checking for a State change so loosely and have other checks to see if the ID of the video has changed or more.
So in my generic example I provided nothing else would break.

src/main/integrations/discord-presence/index.ts
Alipoodle: Please do not resolve this without at least commenting?

Akiisqt: I changed the name of the function back

src/main/integrations/discord-presence/index.ts
Akiisqt: I added checks for the video details id changing, or progress changing, or state changing to update activity.

src/main/integrations/discord-presence/index.ts
Alipoodle: https://github.com/Akiisqt/ytmdesktop/commit/6439a53aaa0178a610091fae90a2efbea166d91a - You fixed it
https://github.com/Akiisqt/ytmdesktop/commit/81cbd1e75b0cd45e6a5edb6a40d880717fb4370e - You reversed it back to the **getBestThumbnail** function... 😖

Akiisqt: see that makes sense now

Sebasee009:
Merge developer into this branch, test with featuring in song name fix.

Akiisqt:
> Merge developer into this branch, test with featuring in song name fix.

?

Alipoodle:
> Merge developer into this branch, test with featuring in song name fix.

This still needs reviewing before it's being merged into the main branch? 😅
Please wait for this PR to be finalised, it will come into the Dev branch soon once it's ready.

src/renderer/windows/settings/Settings.vue
```diff
@@ -391,10 +394,16 @@ window.ytmd.handleUpdateDownloaded(() => {
               <td>No authorized companions</td>
             </div>
           </YTMDSetting>
-          <YTMDSetting v-model="discordPresenceEnabled" type="checkbox" name="Discord rich presence" @change="settingsChanged" />
-          <div v-if="discordPresenceEnabled && discordPresenceConnectionFailed" class="setting indented">
-            <p class="discord-failure">Discord connection could not be established after 30 attempts</p>
-            <button @click="restartDiscordPresence">Retry</button>
+          <div>
```
Alipoodle: This and https://github.com/ytmdesktop/ytmdesktop/blob/def64d1087a6083b30abbdcc21ce05a5d7639dcc/src/renderer/windows/settings/Settings.vue#L407

Shouldn't be included as there's not a need to contain them, and just being consistent with other settings

src/main/integrations/discord-presence/index.ts
```diff
@@ -78,93 +64,78 @@ function stringLimit(str: string, limit: number, minimum: number) {
...
-              url: `ytmd://play/${state.videoDetails.id}`
-            }
-          ]
-        });
-      }
+        this.videoDetails &&
+        this.videoDetails.id &&
+        (oldState !== this.videoState || Math.abs(oldProgress - this.progress) > 5 || oldId !== this.videoDetails.id)
+      )
+        this.setActivity();
```
Alipoodle: I feel like this might need to be re-done. There's 2 primary things to address here including.

I feel like this doesn't read very well at a initial glance of the code. I struggled to understand with the variable names all closely matching up from the `state.videoDetails` and `this.videoDetails` having the property names (but they do store the same thing)


Secondly, line 136 doing the comparisons, does this only check for a change of 5 seconds? Does me constantly rewinding it 1 second mean that my Discord RPC is going to desync now?
While the old method was a bit more complex I feel it is needed.


This might be something which I play around with directly in your Branch to work out.

Alipoodle (changes requested):
This function and the recent changes discussed in [Discord #🧩Integrations Message](https://canary.discord.com/channels/549660225364557835/721151085087293563/1270215793056944210) I think are the only 2 bits of which really need to get poked at in regards to this PR to be fully ready to be approved.

The changes in the Internal API doesn't give everything and this will need some checking on, but I'd recommend reviewing the message from Novus.

The function otherwise appears to be a step back in how often / how the thing updates. ID + Time update or State update should be good in regards to what is needed to change, but how these apply might be worth reviewing.

Sebasee009:
> > Merge developer into this branch, test with featuring in song name fix.
> 
> This still needs reviewing before it's being merged into the main branch? 😅 Please wait for this PR to be finalised, it will come into the Dev branch soon once it's ready.

Oh, I meant to merge the change from development to this branch. sorry

Akiisqt:
What do you mean?

src/main/integrations/discord-presence/index.ts
Akiisqt: From my own research the state changes whenever you move the progress manually. I have yet to find an instance where checking the id and the video state didnt work. But I believe it could be different for people with different setups

Alipoodle: Do you have a Premium account?

Because one example to try would be downloaded music.

Akiisqt: Yes just tested with a downloaded song, and the state still changes, even when looping

NovusTheory (commented):
Looping back to this PR for general commentary and possible changes especially with some further thought on how things should work.

I more-so recently added a new property to the apps state store called `hasFullMetadata`. Ideally since this PR has waited so long to merge I'd love to just get that tackled in if possible. The purpose of that property is to signal that the video's title, album, thumbnails, and other are "finished" being gathered. Ref: YTM sometimes plays the song before the player bar at the bottom shows the metadata and this is when `hasFullMetadata` will become true.

I think a good compromise between what I was asking before with the state not being guaranteed to change would be sufficiently corrected with the `hasFullMetadata` property as a signal for when to update the activity outside of state changing. I still ask that progress be checked individually as it is now in the event YTM *ever* updates to not change the state when seeking.

After some testing with this PR I think we could further improve how we update the Discord activity by debouncing `setActivity` calls. `setActivity` would need to wait e.g. 1 second before sending the activity update to Discord in the event that within that time frame `setActivity` gets called again so it cancels that timeout and we only send 1 update to Discord.

As my final note I'd like to bring up the setting to allow changing from `Playing` to `Listening` and vice versa. I think at the time this was added it was maybe necessary because Discord broke timestamps on Listening but I don't think this was a good idea to add in anyways as there's no reason to have this setting and it's more irrelevant now that Listening has been fixed and given this is a media player there's no reason to have it say anything other than Listening

Akiisqt:
> Looping back to this PR for general commentary and possible changes especially with some further thought on how things should work.
> 
> I more-so recently added a new property to the apps state store called `hasFullMetadata`. Ideally since this PR has waited so long to merge I'd love to just get that tackled in if possible. The purpose of that property is to signal that the video's title, album, thumbnails, and other are "finished" being gathered. Ref: YTM sometimes plays the song before the player bar at the bottom shows the metadata and this is when `hasFullMetadata` will become true.
> 
> I think a good compromise between what I was asking before with the state not being guaranteed to change would be sufficiently corrected with the `hasFullMetadata` property as a signal for when to update the activity outside of state changing. I still ask that progress be checked individually as it is now in the event YTM _ever_ updates to not change the state when seeking.
> 
> After some testing with this PR I think we could further improve how we update the Discord activity by debouncing `setActivity` calls. `setActivity` would need to wait e.g. 1 second before sending the activity update to Discord in the event that within that time frame `setActivity` gets called again so it cancels that timeout and we only send 1 update to Discord.
> 
> As my final note I'd like to bring up the setting to allow changing from `Playing` to `Listening` and vice versa. I think at the time this was added it was maybe necessary because Discord broke timestamps on Listening but I don't think this was a good idea to add in anyways as there's no reason to have this setting and it's more irrelevant now that Listening has been fixed and given this is a media player there's no reason to have it say anything other than Listening

I can remove the custom setting then, I will look into the hasFullMetadata, from my understanding it only becomes true once a song is loaded, have you tested with looping songs? I haven't gotten a chance to look into the new updates. I will also integrate an denounce as requested, shouldn't be hard since it will just use whatever state it has saved.

Akiisqt:
https://github.com/ytmdesktop/ytmdesktop/pull/1454

PR #1411, opened 2024-08-07
✨ Slack integration

enandr:
I have added Slack Integration to update your status on Slack automatically. You can toggle on or off the Slack integration. You can enter your Slack User Token (which you get from creating and installing a custom Slack app to your workspace) and choose the amount of time until your status will be cleared after you pause the music.


src/main/index.ts
```diff
@@ -1341,6 +1358,9 @@ app.on("ready", async () => {
         if ("settings-discord-rich-presence" in v1Config) {
           store.set("integrations.discordPresenceEnabled", v1Config["settings-discord-rich-presence"]);
         }
+        if ("settings-slack-status" in v1Config) {
+          store.set("integrations.slackEnabled", v1Config["slack-status"]);
+        }
```
Alipoodle: This was not included as a feature in v1, can I ask why this was added?

src/main/integrations/slack/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  private store: Conf<StoreSchema>;
+  private enabled = false;
+  private stateCallback: (event: PlayerState) => void = null;
+  private currentVideoId: string = "";
+  private statusTimeoutId: NodeJS.Timeout;
+  private slackUserToken: string = "";
+
+  private async playerStateChanged(state: PlayerState) {
+    const { secondsTillClearStatus } = await this.store.get("slack");
+    if (state.trackState === 1 && state.videoDetails) {
```
Alipoodle: Please can you use the `VideoState` enum instead of the Hard Coded number.

src/main/integrations/slack/index.ts
```diff
@@ -0,0 +1,81 @@
...
+        } else {
+          this.currentVideoId = state.videoDetails.id;
+          this.updateSlackStatus(false, state.videoDetails);
+        }
+      }
+    }
+  }
+  private updateSlackStatus(clear: boolean, data: VideoDetails) {
+    const message = `Listening To: ${data.title} | ${data.author}`;
+    let postData: { status_text: string; status_emoji: string; status_expiration: number } = {
```
Alipoodle: Feel like there's a cleaning way of doing this Type Checking either with a a new `type.ts` file which includes what the data format is for posting data / recieved Data ETC.

src/shared/store/schema.ts
```diff
@@ -52,6 +53,10 @@ export type StoreSchema = {
     sessionKey: string | null;
     scrobblePercent: number;
   };
+  slack: {
+    slackUserToken: string;
```
Alipoodle, line 57: This is a wider thing than just this file 🤔 
But as this is a "Secret" that this should be Encrypted and Decrypted in the integration.

There would probably be some bits in regards to preventing this being overwritten if left blank in settings 🤷

src/renderer/windows/settings/Settings.vue
```diff
@@ -426,6 +436,29 @@ window.ytmd.handleUpdateDownloaded(() => {
...
+          <YTMDSetting
+            v-if="slackEnabled"
+            v-model="secondsTillClearStatus"
+            class="settings indented"
+            type="range"
+            name="Clear Status Seconds"
+            description="Time until the status is automatically cleared"
+            min="5"
+            max="120"
+            step="5"
```
Alipoodle, line 466: This standard or should we allow a more open range
For Discord's rich presneces we have it hard coded for 60 seconds when paused to hide it? Is it not worth while in doing the same for this?

src/renderer/windows/settings/Settings.vue
```diff
@@ -426,6 +436,29 @@ window.ytmd.handleUpdateDownloaded(() => {
             step="5"
             @change="settingsChanged"
           />
+          <YTMDSetting v-model="slackEnabled" type="checkbox" name="Slack" @change="settingsChanged" />
+          <i v-if="slackEnabled">You must create a Slack app and generate a user token to use slack integration</i>
```
Alipoodle: This feels like it is not needed, but I haven't built or seen the in-app thing to understand how this looks.

A user shouldn't have to be told something like this for a setting. If they are enabling a setting I would hope they would at least have some understanding of what it does, especially considering **Slack** is a bit more of a common thing~

src/main/integrations/slack/index.ts
```diff
@@ -0,0 +1,81 @@
...
+      const currentUnixTimestamp = Math.floor(Date.now() / 1000);
+      // Add 15 minutes (15 * 60 seconds)
+      const newUnixTimestamp = currentUnixTimestamp + 15 * 60;
+      postData = {
+        status_text: message,
+        status_emoji: ":notes:",
+        status_expiration: newUnixTimestamp
+      };
+    }
+    if (!this.slackUserToken) return;
```
Alipoodle: Could this be called earlier to prevent processing of the function earlier?

src/main/integrations/slack/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+  private updateSlackStatus(clear: boolean, data: VideoDetails) {
+    const message = `Listening To: ${data.title} | ${data.author}`;
+    let postData: { status_text: string; status_emoji: string; status_expiration: number } = {
+      status_text: "",
+      status_emoji: "",
+      status_expiration: 0
+    };
+    if (!clear) {
+      const currentUnixTimestamp = Math.floor(Date.now() / 1000);
```
Alipoodle: `currentUnixTimestamp` and `message` feel like they're not really needed as variables if they are only storing stuff for a single usage within the same function?

I would personally just put them directly inside of the `postData`.

src/main/integrations/slack/index.ts
```diff
@@ -0,0 +1,81 @@
...
+          this.updateSlackStatus(false, state.videoDetails);
+        }
+      }
+    }
+  }
+  private updateSlackStatus(clear: boolean, data: VideoDetails) {
+    const message = `Listening To: ${data.title} | ${data.author}`;
+    let postData: { status_text: string; status_emoji: string; status_expiration: number } = {
+      status_text: "",
+      status_emoji: "",
```
Alipoodle: *Idea*
🎵 Emoji?

Alipoodle (changes requested):
😋 A few changes from a quick skim of the code.

src/main/index.ts
enandr: I am still learning your codebase. I saw the others using this so I did as well. You are right that Slack is not a v1 feature. I couldn't though find anything for v2Config. Does it currently exist?

src/main/integrations/slack/index.ts
enandr: This has been updated in my latest push.

src/main/integrations/slack/index.ts
enandr: I have created a new type file and put that type there. that one there was supposed to be temporary that I just forgot to move out. its in my latest push

src/renderer/windows/settings/Settings.vue
enandr: I was thinking about hard coding but I decided against it. I have 60 as the default. As Slack is primarily a business platform, not a gaming one like discord, it would be nice to have a range of options.

src/renderer/windows/settings/Settings.vue
enandr: After second thought, you are right, its not needed. It has been removed in my latest push

src/main/integrations/slack/index.ts
enandr: You are right. the changes have been reflected in my latest push

src/main/integrations/slack/index.ts
enandr: the timestamp can be moved from 2 variable to 1 and message can be removed and used without a variable. this has been reflected in my latest push

enandr: The way slack works though is that you must send empty data to clear the status. any data inside will create a status. I have instead created 2 separate post data variables, one with blank data that clears, one with real data that creates a status

src/shared/store/schema.ts
enandr: It would be better as the user token is a secret and should be encrypted. i will update that in my next push.

What do you mean by "This is a wider thing than just this file"

src/main/integrations/slack/index.ts
enandr: Not a bad idea. I could have a box where you can enter the emoji you wish to use or offer a few options. Right now I use 🎶 hard coded

src/main/index.ts
Alipoodle: This is a set of Migration code from V1 -> V2, and this is a new feature.

This was not a feature which was in the v1 of the app, so it is not required.

src/shared/store/schema.ts
Alipoodle: You need to grab the encrypted value in your `slack/index.ts` and decrypt it.
And Currently we do not have a proper settings input field which supports for `TextEncrypted` and handles the initial Encryption on the Settings page.

src/main/@types/slack.d.ts
```diff
@@ -0,0 +1,2 @@
+type PostType = { status_text: string; status_emoji: string; status_expiration: number };
```
Alipoodle: 1) Can this instead be `export type ...` to achieve both the type Creation + sharing it.
2) is it possible to have some new lines in this, just for readability 😄 
3) Can this be renamed to **SlackPostData** or something.
   1) Prefix it with the Integration name (to not confuse it in Intelisense)
   2) `Type` shouldn't be needed to be included as that is what it's defined as.
   3) PostData is actually what this is being used for.

src/main/integrations/slack/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  private stateCallback: (event: PlayerState) => void = null;
+  private currentVideoId: string = "";
+  private statusTimeoutId: NodeJS.Timeout;
+  private slackUserToken: string = "";
+
+  private async playerStateChanged(state: PlayerState) {
+    if (!this.enabled) return;
+    const { secondsTillClearStatus } = await this.store.get("slack");
+    if (state.trackState === VideoState.Playing && state.videoDetails) {
+      if (!state.adPlaying) {
```
Alipoodle: Any reason to not combine these 2 statements or do some early exits instead.
Feels like it would clean up from having a spaghetti of If statements a bit more. ❤

```js
if (!this.enabled ||
    !state.adPlaying ||
    !state.videoDetails ||
    state.trackState !== VideoState.Playing
 ) return;

File: src/main/integrations/slack/index.ts [hunk 12]
```diff
@@ -0,0 +1,78 @@
...
+    const clearPostData: PostType = {
+      status_text: "",
+      status_emoji: "",
+      status_expiration: 0
+    };
+    const listeningToPostData: PostType = {
+      status_text: `Listening To: ${data.title} | ${data.author}`,
+      status_emoji: ":notes:",
+      status_expiration: currentUnixTimestamp + 15 * 60
+    };
```
Alipoodle: Can this become 1 variable instead?
I would play with the formatting a bit to possibly make it a bit more human readable, but something like this is what I am thinking about.
```js
const postData: PostType = (
   clear
   ? {
      status_text: "",
      status_emoji: "",
      status_expiration: 0
    }
    : {
      status_text: `Listening To: ${data.title} | ${data.author}`,
      status_emoji: ":notes:",
      status_expiration: currentUnixTimestamp + 15 * 60
    }
);

Alipoodle (commented):
2 Comments from the changes you've just made.
and 1 which I forgot to see.

Otherwise there's the setting thing from before, which is a bit more of a task overall. 😅

src/shared/store/schema.ts
enandr: the encryption is done. i found the way to encrypt it and decrypt it also on the settings page so you can view it

src/main/integrations/slack/index.ts
enandr: i like what you sent, but i did something similar

src/main/integrations/slack/index.ts
enandr: its now instead of this, each keys value is set based on clear or not

enandr:
@Alipoodle I would love for this to be a part of YTMD and I can then call myself a contributor and do more. I belive all of your review requests are completed

Issue #1423, opened 2024-08-25
App Title on Mac Looks Out of Place


TechRemarker:
### Consent

- [X] I verified that there is no open/closed issue for the same subject.
- [X] I understand that **YTMDesktop have NO affiliation with Google or YouTube**

### Current Behavior

On Mac the top left of the app has an oversized logo and oversized font saying "YouTube Music Desktop App". It looks quite out of place, as Mac apps rarely ever had the app title in the top of the app by the buttons, since the title will already appear in the menu bar and the play icon also seems redundant since the YT Music appears right beneath it.
![CleanShot 2024-08-25 at 09 49 53@2x](https://github.com/user-attachments/assets/8bb91bac-423f-45b0-8f4f-da7a5853abef)
Also while no need for the title there at least on Mac, it also is in a font and size too large for the space making it look very out of place on the app. This goes for the icons in the top right as well, which are too large for the space. There should be more padding around them. 

### Expected Behavior

Follow Mac UI standards on Mac.

### Steps To Reproduce

1. Open App

### YTMDesktop

2.0.5

### OS

Mac

### OS Version

Sequoia

### Arch

ARM64

### Installation way

Mac Download

### Anything else?

![CleanShot 2024-08-25 at 09 49 53@2x](https://github.com/user-attachments/assets/df1724be-9f88-4af2-ad9a-baf841a66eb2)

TechRemarker:
Forgot to add another visual bug where everything is true black except behind the three dots on mac as you can see it's a different shade of black.

PR #1424, opened 2024-08-25
💄 Improve consistency of title-bar (windows)

XeroxDev:
# Summary
This pull request enhances the title-bar to make it a bit more consistent on windows devices. It's just a small change, but IMO looks better (we've also discussed about this in discord).

# What did exactly change?
I adjusted
- the font size of the "YouTube Music Desktop App" Title
- the main icon, so it's not as huge and "out of place"
- the right icons
  - I made the icons self smaller, so it's kinda the same size as the "official" windows buttons
  - I made the buttons (clickable space) somewhat the same size, as the "official" windows buttons
  - I removed the divider, so everything has somewhat the same distance to another (at least for the eyes)

# Comparison
**Before**
image

**After**
image

# Was that needed?
Not necessarily, but I think it's looking more consistent now and less "out of place". But that's just my opinion, if you think the original one is better, feel free to decline and close this pull request

# PS
I've also added `.idea` to the .gitignore. This is the JetBrains IDE settings folder.

# Addressed issues
This also somewhat addresses #1423 but not entirely (mac specific things weren't done, so it's just the icons on the right side, mentioned in the issue)

### DIFF
--- a/src/renderer/components/TitleBar.vue
+++ b/src/renderer/components/TitleBar.vue
@@ -147,6 +147,7 @@ if (props.isMainWindow) {
 .titlebar .right .app-buttons {
   display: flex;
   flex-direction: row;
+  margin-right: 16px;
 }
 
 .title {
@@ -184,8 +185,9 @@ if (props.isMainWindow) {
 }
 
 .app-button {
-  height: 35px;
-  width: 44px;
+  margin-right: 4px;
+  height: 28px;
+  width: 28px;
   background: none;
   color: #bbbbbb;
   display: flex;
@@ -193,7 +195,7 @@ if (props.isMainWindow) {
   justify-content: center;
   -webkit-app-region: no-drag;
   border: none;
-  border-radius: 0;
+  border-radius: 4px;
   font-variation-settings:
     "FILL" 0,
     "wght" 200,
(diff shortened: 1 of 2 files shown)


src/renderer/components/TitleBar.vue
```diff
@@ -158,54 +156,59 @@ if (props.isMainWindow) {
...
+.title .icon.material-symbols-outlined {
+  font-size: 18px;
 }
 
 .title-text {
   font-family: "Open Sans", sans-serif;
   min-width: 0;
   white-space: nowrap;
   overflow: hidden;
+  font-size: 13px;
```
NovusTheory: Having tested this seems like it could be 14px as 13px seems 1 pixel too small

NovusTheory (changes requested):
Overall looks good besides a couple nitpicks.

An overall change I'm not sure I agree with is the app buttons we add to the titlebar matching Windows explicitly. This does affect other platforms so they make look off. I'm open to changing the margin/padding but I'd rather have it look more like before where it aligns with what IntelliJ IDEA or VSCode does for extra top right titlebar buttons.

XeroxDev:
As requested and discussed on discord, I've changed the top right icons a bit
image

NovusTheory (approved):

PR #1454, opened 2024-10-24
🎨✨🩹 Discord Presence Improvements

Akiisqt:
**Optimized Code:**
 Improved readability.
 Removed excess or unnecessary code.
 Improved performance.

**Activity Management:**
 Removed the redundant setting for _ActivityType._
 Now utilizing _hasFullMetadata_ for updating activity.
 Added a debounce to reduce the frequency of updates sent to Discord.

### DIFF
--- a/src/main/integrations/discord-presence/index.ts
+++ b/src/main/integrations/discord-presence/index.ts
@@ -1,4 +1,4 @@
-import playerStateStore, { PlayerState, Thumbnail, VideoState } from "../../player-state-store";
+import playerStateStore, { PlayerState, Thumbnail, VideoState, VideoDetails } from "../../player-state-store";
 import IIntegration from "../integration";
 import MemoryStore from "../../memory-store";
 import { MemoryStoreSchema } from "~shared/store/schema";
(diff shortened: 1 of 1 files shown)


Akiisqt:
Not sure how I feel about the new function that replaces apart of the if statement, Im open if you guys believe it should just be the if statement and not have some moved to a separate function

src/main/integrations/discord-presence/index.ts
```diff
@@ -82,172 +55,146 @@ export default class DiscordPresence implements IIntegration {
...
-          adjustedTimeSeconds = this.lastTimeSeconds;
-        } else {
-          this.lastDuration = duration;
-          this.lastTimeSeconds = timeSeconds;
-        }
-      } else {
-        this.lastDuration = duration;
-        this.lastTimeSeconds = timeSeconds;
+  private setActivity() {
+    if (this.activityDebounceTimeout) return;
```
NovusTheory, line 70: This should clear the timeout instead of returning. Currently this would only take old data when setting the activity if it were to be called multiple times within one second (as noted by how long the setTimeout is) and the latest call wouldn't be used.

NovusTheory (changes requested):
Just one small change requested and LGTM after that from an initial look over

src/main/integrations/discord-presence/index.ts
Akiisqt: It is better like this since it will send the request to update and will update in 1 second.

But it will not delay it more than 1 second. If it cancelled it right away then it could be forever delayed. This method allows for it to be sent within 1 second and not potentially 10 minutes later, but extreme but yk

NovusTheory: That's not at all what it would be doing. setActivity would be creating a new timeout to replace the one you just cleared so I'm not sure where you're getting the forever delayed logic from. The point of this is so that the **latest** call to setActivity is the data used but your current logic allows setActivity to be called 100 times within 1 second but only the first call's activity data is used where as the 100th call should be the actual activity being set.

Akiisqt: It is the 100th's data being sent, setActivity uses globals

NovusTheory: Ah that makes sense now. I got a bit confused on the logic there. I think the function should be changed from `setActivity` to `updateActivity`. Using `set` kind of implies we're passing data to that function to be set but that's not what's logically happening.

Akiisqt: 👍

Akiisqt (commented):
Lgtm

Akiisqt:
Will fix https://github.com/ytmdesktop/ytmdesktop/issues/1417

NovusTheory (approved):

PR #1461, opened 2024-11-02
Improve errors on invalid api requests by being less strict in schema

Setadokalo:
When calling the companion server API with incorrect parameters, it often produces unpleasant and hard to interpret error messages. For example, for the request `{ "command": "setVolume", "data": 101 }`, the API will respond:

```{
  "statusCode": 400,
  "code": "FST_ERR_VALIDATION",
  "error": "Bad Request",
  "message": "body/command must be equal to constant, body/command must be equal to constant, body/command must be equal to constant, body/command must be equal to constant, body/command must be equal to constant, body/data must be <= 100, body/command must be equal to constant, body/command must be equal to constant, body/command must be equal to constant, body/command must be equal to constant, body/command must be equal to constant, body/command must be equal to constant, body/command must be equal to constant, body/command must be equal to constant, body/command must be equal to constant, body/command must be equal to constant, body/command must be equal to constant, body must match a schema in anyOf"
}
```

(This was formatted by the `reqwest` crate, but is still just representing the raw http response from the API.)

This is ultimately due to the error being produced by the fastify schema parser and listing every possible command and where in the parsing it failed (with all but one failing due to the `command` literal being that of a different command). By providing a custom schema error formatter in `companion-server/index.ts`, we can filter out most of those errors, but the rest still aren't very descriptive when `data` is out of range. So instead, we can just drop the range requirements from the schema, and let the callbacks for each command handle the range validation instead. In fact, all but one of them was *already* checking the range, despite it being validated by the schema.

With this change, the above response becomes:

```
{
  "statusCode": 400,
  "code": "INVALID_VOLUME",
  "error": "Bad Request",
  "message": "Volume '101' is invalid"
}
```
Additionally, requests with invalid commands become:
```
{
  "statusCode": 400,
  "code": "FST_ERR_VALIDATION",
  "error": "Bad Request",
  "message": "invalid command"
}
```
And commands with a missing `data` parameter become:

```
{
  "statusCode": 400,
  "code": "FST_ERR_VALIDATION",
  "error": "Bad Request",
  "message": "must have required property 'data'"
}
```

### DIFF
--- a/src/main/integrations/companion-server/api-shared/schemas.ts
+++ b/src/main/integrations/companion-server/api-shared/schemas.ts
@@ -19,10 +19,7 @@ export const APIV1CommandRequestBody = Type.Union([
   }),
   Type.Object({
     command: Type.Literal("setVolume"),
-    data: Type.Number({
-      minimum: 0,
-      maximum: 100
-    })
+    data: Type.Number()
   }),
   Type.Object({
     command: Type.Literal("mute")
@@ -32,9 +29,7 @@ export const APIV1CommandRequestBody = Type.Union([
   }),
   Type.Object({
     command: Type.Literal("seekTo"),
-    data: Type.Number({
-      minimum: 0
-    })
+    data: Type.Number()
   }),
   Type.Object({
     command: Type.Literal("next")
@@ -51,9 +46,7 @@ export const APIV1CommandRequestBody = Type.Union([
   }),
   Type.Object({
     command: Type.Literal("playQueueIndex"),
-    data: Type.Number({
-      minimum: 0
-    })
+    data: Type.Number()
   }),
   Type.Object({
     command: Type.Literal("changeVideo"),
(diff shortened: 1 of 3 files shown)


NovusTheory:
I definitely agree that error is very harsh but I'm not quite sure I agree with removing those validations out of the schema and just validating them inside of the command itself. It sort of defeats the purpose of the schema being there. I do understand that *some* stuff was not in the schema and it was mostly my laziness and some data being dynamic based on the state (e.g. playQueueIndex) and I didn't at the time want to put effort to figuring out how to tie it more to the schema.

Setadokalo:
My initial plan was just to clean up the schema error and make it clearer, but the formatter doesn't seem to get enough information about the error to clearly explain what's wrong without duplicating the schema order and counting errors to figure out which command actually failed to parse. Reading the docs again though, it might be possible to set the error message to be more clear in the schema.

That said, is there an advantage to doing these limits in the schema? It doesn't look like dynamic range limits are supported, so no matter what at least some of these ranges have to be checked in the callback, and most commands were checking the ranges anyways.

NovusTheory:
> That said, is there an advantage to doing these limits in the schema?

Code organization and centralizing the handling of schema definition and validation so the request function just has to deal with data that is guaranteed valid.

PR #1474, opened 2024-12-04
Allow to choose Dark/Light tray icon

### LINKED ISSUE #1356: Tray icon for dark mode on Linux
### Consent

### Feature

I am using Ubuntu and my tray/panel background is black. When YTM hide to tray, the icon is barely visible. Could we somehow bring back the old red icon in version 1? It should be ok on any background
Current icon
image

Old icon is something like this
image

### Solution

_No response_

### Alternatives

_No response_

### Anything else?

_No response_

hron:
Some quick demo:

https://github.com/user-attachments/assets/183c3d54-a534-4268-94a2-77922941eda5

Closes #1356


Alipoodle:
_this is without testing and checking docs_
But trayTemplate was a specific name required for MacOS to automatically do it on that OS~
So this maybe something which breaks that functionality ~


Edit:
Details can be found here: https://www.electronjs.org/docs/latest/api/native-image#template-image-macos

Edit 2:
Additionally this feels like it would be nice to also include Windows and have a Inverted .ico file as well.

hron:
@Alipoodle Thank you for the quick feedback! I trust you about macOS, but I think I have found a universal way to handle light/dark tray icon style which can be utilized everywhere.  https://www.electronjs.org/docs/latest/api/native-theme 

I cannot test it on macOS, but it works on Linux. The new version goes one step further and introduces a setting to either do light/dark switching automatically depending on 'Dark Mode' setting or allow the user to choose. This is useful for situation when a light theme is in use, but the panel with tray is dark. 

The updated version should work on Windows as well as you requested. Please let me know if something is missing. Also, here is new demo:

https://github.com/user-attachments/assets/672c4e47-6a98-4c96-81b8-f51b9a0ffed5

NovusTheory:
> Additionally this feels like it would be nice to also include Windows and have a Inverted .ico file as well.

The normal for Windows is usually to just use the app's colored icon and not some specialized tray icon which can flip colors like this.

While this PR is only to toggle dark/light icons I feel like if we'll be messing with the tray icon we should just try and swap to something better like maybe utilizing our black and white logos that are already in the repo (we already use the white for the titlebar). Mac is still a special case though as UI and designs there are drastically different

hron:
@NovusTheory all your feedback has been incorporated. In addition, the setting to change the tray icon style is only shown on Linux now. I think it makes sense because on macOS the app uses trayTemplate.png and on Windows it always shows the colored version of the tray icon, so the setting doesn't make any sense there.

Demo:


https://github.com/user-attachments/assets/09616614-494d-4d97-bc2c-df524735869b

NovusTheory:
Outside of the CSS for the select menu missing and not aligning with the rest of the app's style I'm satisfied with these changes.

hron:
Well, CSS is not my strong side, but here are some improvements:


https://github.com/user-attachments/assets/e30d5ee6-c518-4be1-9745-a41c0934ca0e

NovusTheory:
I had just got reminded of an old branch I made for something else which implemented a select menu setting https://github.com/ytmdesktop/ytmdesktop/commit/363916b3418e1e30b095ef2edc6aeb2118bc246b. After looking over the code for this PR I think I'd prefer the logic in the linked commit for our setting component. I can cherry-pick that select menu code for our setting component (YTMDSetting.vue) into the development branch and this PR could get adjusted to utilize that if you'd be willing to. If that branch had ever gotten merged then it would've already been available and ready to use for this PR.

hron:
@NovusTheory sounds good to me, please ping me when the select implementation is available in `development`, so I can rebase this branch and use the new component. Thanks!

NovusTheory:
@hron Pushed the select menu type to development and ready for use.

hron:
@NovusTheory okay, it's done and ready for another review round

src/renderer/windows/settings/Settings.vue
```diff
@@ -313,6 +326,14 @@ window.ytmd.handleUpdateDownloaded(() => {
             @clear="removeCustomCSSPath"
           />
           <YTMDSetting v-model="zoom" type="range" max="300" min="30" step="10" name="Zoom" @change="settingsChanged" />
+          <YTMDSetting
+            v-if="isLinux"
+            v-model="trayIconStyle"
+            :options-map="toOptionsMap(trayIconStyles)"
```
NovusTheory: Nit: This should capitalize the options for the UI. The actual config doesn't need it.

NovusTheory (changes requested):
Not sure if it would be beneficial to make the config an enum (number) instead of arbitrary text but otherwise seems okay outside a nit change.

@Alipoodle Chiming you in for thoughts on the above.

hron:
> Not sure if it would be beneficial to make the config an enum (number) instead of arbitrary text but otherwise seems okay outside a nit change.
> 
> @Alipoodle Chiming you in for thoughts on the above.

I have updated the codebase to use an enum (number). It definitely looks simpler on the code side, so I believe this change makes sense. Although, I had to change YTMSetting component a bit. It was working a bit incorrectly because when an option is selected the value it used for assignment was a `string` (e.g. "1"), not a `number` (e.g. 1).

Also, the options are capitalized now as requested:

image

hron:
@NovusTheory any more comments on this? I think the PR is ready for merging

NovusTheory (approved):
Sorry this took so long to get around to. LGTM

PR #1515, opened 2025-03-25
Update Settings.vue

Frikelangelo:
Line 368 change= "memorysetting..." to just change="settings..." So i can open my YT Desktop with Enable Companion Authorization always on instead of check it each time i open it... cause is a bit frustating

### DIFF
--- a/src/renderer/windows/settings/Settings.vue
+++ b/src/renderer/windows/settings/Settings.vue
@@ -365,7 +365,7 @@ window.ytmd.handleUpdateDownloaded(() => {
             indented
             name="Enable companion authorization"
             description="Automatically disables after the first successful authorization or 5 minutes has passed"
-            @change="memorySettingsChanged"
+            @change="settingsChanged"
           />
           <YTMDSetting
             v-if="companionServerEnabled && safeStorageAvailable"


Alipoodle:
I'm confused to this change? The intended purpose of this toggle is to temporarily allow for connections to be allowed in (for 5 minutes)

It sounds like your application which is connecting to YTM Desktop isn't correctly storing it's token to allow for it to work without the need of enabling this settings again?

Frikelangelo:
The option "Enable companion authorization" give me permission to connect it and get control, by not only device also other applications. For example when im coding in VS, or gaming (example MC) where i connect to the YT Desktop App but when i turn off PC and run it again. Basically all days, maybe few times in a day ( Work and entertainment in same place issues ) The memorySetting Store on Volatile parts so i have to enable it again and again and again each time. I'm not sure if it also happen by just closing YT Desktop APP. Think not but not sure. So it would be amazing to be the option to Permantly ENABLE or activate the Toggle Memory (Register Option). Just so when i run the app the first 5 minits it got my ENABLE config, it connects automatically without me needing to go inside options and check it again.

Sorry I've said something confusing or by my english, i think i got not bad level but is not my mother language... so any doubt just ask me and I will try to explain better.

PD: TY for answering and reading my message <3

Alipoodle:
This sounds like a problem in the VS Code plugin/Minecraft mod not storing the token used to connect. 

You should only need to enable this on a new application connecting to YouTube music. Existing connections can use their token to renew it after its no longer valid and not require this setting to be re-enabled.

Alipoodle:
Could you share the VS plugin and Minecraft mod you are using?

Frikelangelo:
Yes, ofc. 
Maybe u r right and its problem for the others linkers, but what i told u i know it could solve it. And I can not solve for example the MC Mod myself. So i finally try to come here and talk to u. 
Not gonna lie i didnt expect any answer to my petition, so thank you for take your time to listen to a Junior :). Im so gratefull for it.

MC Mod Craftify - https://modrinth.com/mod/craftify/versions  
(I'm using last one but I have already used others versions and it also got the same problem. The thing is that I only need to open YT App and click the botton to solve it, the "token used to connect" is saved correctly and didnt change it never... in the MC Mod Config so cause this I began checking this app on the github and found this thing, that I think it could solve it. Maybe Im just a dreamer) 😂

VS Code
https://github.com/sedders123/vscode-ytmusic/releases/tag/v2.0.0

sedders123:
`vscode-ytmusic` should be storing and using the auth token, if you're experiencing different behaviour please do [open an issue](https://github.com/sedders123/vscode-ytmusic/issues/new) and I can try and take a look.

Issue #1542, opened 2025-06-05
Shuffle state in Companion API


MaverickMartyn:
### Consent

- [x] I verified that there is no open/closed issue for the same subject.
- [x] I understand that **YTMDesktop have NO affiliation with Google or YouTube**

### Feature

I'm using the companion API to integrate with ytmdesktop, and I can't seem to find any information in the state regarding the shuffle setting.  
From the documentation, I can tell that it's possible to set it, but I can't see any way to retrieve the current value.

### Solution

Add shuffle information to the state exposed in the Companion API.

### Alternatives

I see none.

Akiisqt:
nevermind

PR #1469, opened 2024-11-18
WIP: v2.1 (prev: Architecture Rework)

NovusTheory:
Edit: This PR used to be just about an architecture rework but now it's v2.1

This is a complete internal architecture and codebase rework which helps improve the readability and functionality of the app from the initial v2 version that had more of a MVP codebase. As more and more gets added it becomes more obvious that this change needs to happen sooner than later to ensure the codebase doesn't become a mess.

### If you are a developer contributing to YTMDesktop this PR will affect any current work as it is a very heavy codebase change among implementing tons of other features

Notable changes
- Singleton service style architecture
   - Watchdog
       - Crash and log handling
   - Auto updater
       - Also changes the app to open an updater window on startup just for update checking before launching the main program
   - Config store
       - Everything inside config.json
   - State manager
       - Handles the current page url and window bounds to flush them to disk more frequently
   - Integration manager
       - Introduces lifecycle runs to integrations and a base class for more ease of programming them
   - YTM view manager
       - Handles everything related to showing YouTube Music
       - Other systems to show YTM regardless if the app can hook it or not
           - A warning dialog is shown when hook errors happen to alert the user functionality is degraded but not limiting usage of YTM entirely
   - Window manager
       - Handles all windows and views in the app and can automatically recreate them if they're closed abnormally e.g. crashed
       - Manages automatic sizing for views attached to windows
   - Protocol manager
       - Manages the ytmd:// protocol
       - Manages the internal ytmd-app:// protocol which is used to serve the apps files and [follows Electrons best practices for security](https://www.electronjs.org/docs/latest/tutorial/security#18-avoid-usage-of-the-file-protocol-and-prefer-usage-of-custom-protocols)

Feature additions
- Timed Lyrics
- shuffleEnabled state in the companion API
- Queue modification commands in the companion API
- Read-only IPC server for local apps to read the state of YTM
- Window title now displays `Video Title - Author` and more consistently
- Updated discord rich presence features
- Enhanced media service via Xosms
- Miniplayer
- mDNS broadcasting (_ytmdesktop._tcp) to find active instances of a companion server on the network. Useful for mobile app developers or anything off the same system as YTMD

PR #1545, opened 2025-06-10
Added shuffle information to state.

MaverickMartyn:
I added the current shuffle state to the reported state information.
See issue #1542

### DIFF
--- a/src/renderer/ytmview/scripts/hookplayerapievents.script.js
+++ b/src/renderer/ytmview/scripts/hookplayerapievents.script.js
@@ -14,8 +14,9 @@
     const volume = state.player.volume;
     const adPlaying = state.player.adPlaying;
     const muted = state.player.muted;
+    const shuffled = !!document.querySelector("ytmusic-app-layout>ytmusic-player-bar").querySelector("ytmusic-player-bar[shuffle-on] .shuffle.ytmusic-player-bar");
 
-    window.ytmd.sendStoreUpdate(state.queue, likeStatus, volume, muted, adPlaying);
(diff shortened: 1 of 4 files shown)


src/renderer/ytmview/scripts/hookplayerapievents.script.js
```diff
@@ -14,8 +14,9 @@
     const volume = state.player.volume;
     const adPlaying = state.player.adPlaying;
     const muted = state.player.muted;
+    const shuffled = !!document.querySelector("ytmusic-app-layout>ytmusic-player-bar").querySelector("ytmusic-player-bar[shuffle-on] .shuffle.ytmusic-player-bar");
```
NovusTheory, line 17: Can we swap this to a state store property like the rest of the information about the state. We avoid doing any DOM querying and rely on YTMs internal state store to provide us information.

NovusTheory (changes requested):
Since the linked issue states Companion API exposure this PR currently does not expose the shuffled property as the API does its own transformation and selectively picks things out of this apps internal state store. I imagine this is something wanted in the PR as well?

Akiisqt:
Please add this to the issue: https://github.com/ytmdesktop/ytmdesktop/issues/1542 https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

src/renderer/ytmview/scripts/hookplayerapievents.script.js
xCSxXenon: Not to expose my lack of knowledge, but isn't that as simple as turning this line into "const shuffled = state.player.shuffled;"

NovusTheory:
Closing in favor of #1469 which implements this

PR #1568, opened 2025-08-06
Adding shuffle state to player states and companion API

xCSxXenon:
#1545  went stale so I made the changes to address #1542 and the request by @NovusTheory

### DIFF
--- a/src/main/integrations/companion-server/api/v1/index.ts
+++ b/src/main/integrations/companion-server/api/v1/index.ts
@@ -40,6 +40,7 @@ const transformPlayerState = (state: PlayerState) => {
       videoProgress: state.videoProgress,
       volume: state.volume,
       muted: state.muted,
+      shuffled: state.shuffled,
       adPlaying: state.adPlaying,
       queue: state.queue
         ? {
--- a/src/renderer/ytmview/scripts/hookplayerapievents.script.js
+++ b/src/renderer/ytmview/scripts/hookplayerapievents.script.js
@@ -14,8 +14,9 @@
     const volume = state.player.volume;
     const adPlaying = state.player.adPlaying;
     const muted = state.player.muted;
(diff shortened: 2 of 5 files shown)


MaverickMartyn:
Awesome. 😃   
I ended up focusing on a different feature in the app I'm integrating, so the PR went stale.  
Hopefully this will be released by the time I return to that integration. 🤞

src/renderer/ytmview/scripts/hookplayerapievents.script.js
```diff
@@ -14,8 +14,9 @@
     const volume = state.player.volume;
     const adPlaying = state.player.adPlaying;
     const muted = state.player.muted;
+    const shuffled = state.player.shuffled;
```
NovusTheory, line 17: This does not seem to be where shuffle status is. Digging into this myself it's part of the queue object which we already send entirely through IPC to the app. It's called `shuffleEnabled`.

src/main/integrations/companion-server/api/v1/index.ts
```diff
@@ -40,6 +40,7 @@ const transformPlayerState = (state: PlayerState) => {
       videoProgress: state.videoProgress,
       volume: state.volume,
       muted: state.muted,
+      shuffled: state.shuffled,
```
NovusTheory, line 43: Considering where the shuffle state actually is I would amend it to the queue object we send right below

NovusTheory (changes requested):
Appreciate that changes were made to satisfy the concerns but please test your changes as it would reveal this was not a valid property to query for. Maybe not that big of a deal but it would've been nice if you didn't overwrite the commit author from the original stale PR so we can keep credit where due

poveu:
I have opened a new PR with correct implementation: https://github.com/ytmdesktop/ytmdesktop/pull/1573

Issue #1589, opened 2025-09-22
Rich Presence is "frozen" on one song, not matched with what's actually playing


Snowbloxian:
### Consent

- [x] I verified that there is no open/closed issue for the same subject.
- [x] I understand that **YTMDesktop have NO affiliation with Google or YouTube**

### Current Behavior

<img width="448" height="162" alt="Image" src="https://github.com/user-attachments/assets/9787bf60-08a7-4b4e-8969-3fb5f5c30c70" />
Rich presence is frozen at the end of a song, while it says "Playing"
Above is what the rich presence says, below is whats actually playing at the time

<img width="1356" height="1082" alt="Image" src="https://github.com/user-attachments/assets/fbeb3162-ba95-4cbf-8b9f-6b8b9a76e5f3" />

### Expected Behavior

Expected behavior is when the rich presence is matching with the current song and progress

### Steps To Reproduce

I cannot explain this, I have no clue what's happening or how I did this

### YTMDesktop

2.0.10

### OS

Windows

### OS Version

10

### Arch

x64

### Installation way

Installed through the installer downloaded on your website (https://ytmdesktop.app)

Snowbloxian:
Update: I was able to fix this by turning off the rich presence then turning it back on again, but this is still an odd bug

Snowbloxian:
> Update: I was able to fix this by turning off the rich presence then turning it back on again, but this is still an odd bug

nevermind its back

Akiisqt:
I was able to recreate this issue by spamming playing songs and pausing.

<img width="1023" height="432" alt="Image" src="https://github.com/user-attachments/assets/e6cc0789-1995-478d-9474-3c9d1c691fdc" />

Alipoodle:
> I was able to recreate this issue by spamming playing songs and pausing.
> 
[...]

PR #1603, opened 2025-10-27
DiscordPresence fixes

Akiisqt:
Implements fix for #1589


Akiisqt:
Did some testing, seemed to fix all the issues I found, such as songs not updating or when you turn presence on or off, and it's not showing the current song.

Akiisqt:
@Alipoodle Unsure if this is the best solution, maybe you could come up with something better...

PR #1616, opened 2025-11-13
Migration from BrowserView to WebContentsView

jamespullar:
https://www.electronjs.org/blog/migrate-to-webcontentsview

[`BrowserView`](https://www.electronjs.org/docs/latest/api/browser-view) was officially deprecated in favor of [`WebContentsView`](https://www.electronjs.org/docs/latest/api/web-contents-view) with the [release of Electron 30](https://www.electronjs.org/blog/electron-30-0). 

## Notes
- Setting background color explicitly to transparent as `WebContentsView` defaults to a white background. `BrowserView` defaulted to a transparent background. This may not be completely necessary considering how styling is applied, but may prevent any potential white screen flashes as the page initially loads.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1,7 +1,7 @@
 import {
   app,
   autoUpdater,
-  BrowserView,
+  WebContentsView,
   BrowserWindow,
   clipboard,
   crashReporter,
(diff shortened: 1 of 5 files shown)


Alipoodle:
This PR likely isn't going to be required as #1469 already exists to fix this and a lot more internal changes.

jamespullar:
Thanks for mentioning that! I'll make sure to familiarize myself with that changeset before I attempt anything else :D

Alipoodle:
> Thanks for mentioning that! I'll make sure to familiarize myself with that changeset before I attempt anything else :D

That PR especially makes a lot of internal changes so this is likely to make making anything new on the Current codebase a little difficult~

PR #1636, opened 2025-12-24
Add configurable volume delta for shortcuts

rzmeu:
Configurable volume delta for shortcuts - the current value is hardcoded to 10% which does not allow precise control. Added a slider which allows to select between 1-20%.

image


makethegod:
Yes Please add this! I want smoother control of the volume and I swear you used to be able to do this in the setting file, but no more...

rzmeu:
@makethegod I don't know if someone is looking at PR's and I am not using this app anymore. I created my own using WebView in WinForms fully customised for my needs. Turns out it's quite easy to make.

PR #1661, opened 2026-02-12
fix(ytmview): resolve YouTube Music loading and rendering issues

milesbuckton:
## Summary
This PR fixes two critical issues that prevented YouTube Music from loading properly in the desktop application.

### 1. BrowserView Rendering Bug (Primary Issue)
The YouTube Music view was attempting to load URLs before being attached to the main window, which prevented proper rendering. The page would appear blank until DevTools was opened, which forced a repaint.

**Solution:** Moved `mainWindow.addBrowserView(ytmView)` to execute BEFORE any `loadURL()` calls, ensuring the view is properly attached and can render when the page loads.

**Location:** `src/main/index.ts` lines 1178-1188

### 2. Preload Script Timeout Handling
Added robust error handling and timeouts to prevent the app from hanging indefinitely if YouTube Music's internal components fail to initialize.

**Changes:**
- Added 15-second timeout for YouTube Music app hook initialization
- Added 15-second timeout for player API readiness check  
- Wrapped all UI initialization steps in try-catch blocks for graceful degradation
- Added error event handler for Material Symbols font loading
- Added comprehensive error logging for debugging

**Location:** `src/renderer/ytmview/preload.ts`

### 3. Additional Improvements
- Added null checks for DOM elements before manipulation
- Made volume slider setup more defensive
- Improved error messages for troubleshooting
- Added debug batch file for easier development on Windows

## Testing
- ✅ YouTube Music loads reliably without timeout errors
- ✅ No longer requires DevTools to be open for the page to render
- ✅ Graceful fallback when individual components fail to initialize
- ✅ Tested on Windows 11 with Electron 38.0.0

## Additional Changes
- Added `package-lock.json` to `.gitignore` (project uses Yarn)
- Added `start-debug.bat` for easier development debugging on Windows

## Related Issues
Fixes the "YouTube Music is taking longer than usual to load" timeout error that was occurring due to the BrowserView not being properly attached before page load.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,9 @@ typings/
 # Optional npm cache directory
 .npm
 
+# npm lock file (using yarn instead)
+package-lock.json
+
 # Optional eslint cache
 .eslintcache
 
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -147,7 +147,9 @@ Object.assign(console, log.functions);
 //#endregion  Crash + Error reporting
 
 // Handle creating/removing shortcuts on Windows when installing/uninstalling.
+log.info("Checking electron squirrel startup...");
 if (electronSquirrelStartup) {
+  log.info("Squirrel startup detected, quitting...");
   app.quit();
 }
 
@@ -1175,6 +1177,18 @@ const createYTMView = (): void => {
 
   memoryStore.set("ytmViewLoadingStatus", "Initialized");
 
+  // Attach the BrowserView to the window BEFORE loading any URL
+  // This is critical to ensure proper rendering
+  if (mainWindow) {
+    mainWindow.addBrowserView(ytmView);
+    ytmView.setBounds({
+      x: 0,
+      y: 36,
+      width: mainWindow.getContentBounds().width,
+      height: mainWindow.getContentBounds().height - 36
+    });
+  }
+
   let navigateDefault = true;
 
   const continueWhereYouLeftOff: boolean = store.get("playback.continueWhereYouLeftOff");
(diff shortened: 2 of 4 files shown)


Alipoodle:
I take it that this isn't really a "resolve" and more of a Avoid the error 👀?

Additionally please note that teh `start-debug.bat` is not really recommended.
We personally would prefer if people were to use the existing [VS Code Launch](https://github.com/ytmdesktop/ytmdesktop/blob/development/.vscode/launch.json) which will properly Launch the application with debugging attached and log the stuff into the Log section of the window.

If you have other development tools which might handle this as well, then we will add them, but a generic windows onne feels a bit unneeded.

milesbuckton:
Hello @Alipoodle, thank you so much for taking the time to code review my PR. I've reverted the commit where I added the `start-debug.bat` file. I do get your point, and I also feel that the VS Code Launch works much better. I was actually in two minds as to whether I should commit the file, or not.

Do you have any pointers on how we can resolve this issue if you think my fix only addresses the symptoms, and not the root cause? I really want to resolve this issue once and for all! The current "fix" does work on my Windows machine, but I'm not sure about the other platforms.

Alipoodle:
> Do you have any pointers on how we can resolve this issue if you think my fix only addresses the symptoms, and not the root cause? I really want to resolve this issue once and for all! The current "fix" does work on my Windows machine, but I'm not sure about the other platforms.

The overall issue is fixed already in v2.0.11?
Being honest, I suspect that most of this is just AI Hallucination in trying to fix isues which don't exist.

There's some aspects which I do agree could be used to help improve stuff
```diff
(function() {
-   return document.querySelector("ytmusic-app-layout>ytmusic-player-bar").playerApi.isReady();
+   const playerBar = document.querySelector("ytmusic-app-layout>ytmusic-player-bar");
+   return playerBar && playerBar.playerApi && playerBar.playerApi.isReady();
})
```
To my knowledge some of these changes won't be required as much due to an overall change to architechure which will prompt the user if it fails if they wish to continue but some features might not work.

milesbuckton:
It looks like `v2.0.11` is working for me now. 🙌🏻 Thank you @Alipoodle. I appreciate your input. 😄 I'm not sure why it wasn't working previously. 🤷🏻‍♂️ Perhaps it was because I was using `v2.0.10`. I'll close this PR as it is no longer relevant, or necessary.

Issue #809, opened 2021-09-05
Realtime Lyrics


Nuyube:
- [x] I understand that **YTMDesktop have NO affiliation with Google or YouTube**.
- [x] I verified that there is no open issue for the same subject.

**Is your feature request related to a problem? Please describe.**
One of the things I hate the most about Youtube Music is that there isn't any realtime lyric support. It seems that TIDAL supports realtime lyrics from Musixmatch. It seems that Spotify recently did the same, but it would be nice to have that ability in Youtube Music.

**Describe the solution you'd like**
It would be cool if the Lyrics window could be updated to use Musixmatch realtime lyrics data (autoscroll, highlight current line)

**Describe alternatives you've considered**
I've considered using TIDAL but several of my favorite artists are simply not on the site.

**Additional context**
Nothing to add.

Alipoodle:
I should firstly note:
We're a community project and not actually Google/YouTube Music. 
The YTM app I believe already supports this, and implementation maybe added to the desktop app in the future. (I'm not them I don't fully know)

We currently have a Lyric page which can be shown up in the right click menu when the full screen Queue & current song is up.

I doubt that there is a free Solution out there which does provide timestamp for lyric lines, I feel that this maybe difficult to complete.
Although I have not done research into if there are free solutions with the time on each lines. 

Due to something like Musixmatch being a more business focused and paid solution for real time lyrics I don't think this is very feasible solution for a more or less free & community ran program. 😅


Please note: this is my opinion and that you should leave this open as a feature which is still possible, but maybe something which may not be implemented due to some of the reason I've specified until a solution is found.

Afoulky:
It is possible to add the MusixMatch lyrics, look at the Spicetify project. There is a custom app that uses the MusixMatch API for sync lyrics.

To get the lyrics, you can find using this : https://apic-desktop.musixmatch.com/ws/1.1/macro.subtitles.get?format=json&q_track={%track%}&q_artist={%artist%}&app_id=web-desktop-app-v1.0&usertoken=%usertoken%

Replace %track% by track name, %artist% by artist name and %usertoken% by your token.

To get a token, here is a tutorial : https://github.com/khanhas/spicetify-cli/wiki/Musixmatch-Token

You don't even need to pay for an access to their API

Alipoodle:
😅 This really doesn't seem like an Ethical way of accessing their API.
Considering this is accessing the Dev Console of their App to get A API Token.

[...]

PR #1666, opened 2026-02-14
Add synced lyrics feature with native-tab UX

ShorMeneses:
Support for Synced Lyrics directly into the desktop app! Using the same endpoints as the mobile app.

The fallback in case of any error is to hide the new UI section, so we don't impact users in case of an issue.

Example:
image
image

How I reverse-engineered the Android API:

This PR implements a new lyrics provider by reverse-engineering the YouTube Music Android client (v8.09.50). It bypasses standard network protections to fetch both synced and static lyrics directly from the YouTube Music internal API.

### The "How":
<details>
<summary><b>Click to expand: Technical Investigation Details</b></summary>

The Investigation Setup: I installed YouTube Music (com.google.android.apps.youtube.music v8.09.50) on a rooted Android Studio Emulator (Android 13).
Hooking the Network Layers: Since Google apps use Cronet (Chromium's custom network stack) which pins certificates and bypasses traditional HTTP proxies, I used Frida to hook directly into the loaded libraries while the app was running in memory.
Log Extraction via Cronet: I wrote Frida scripts to programmatically modify the Cronet engine parameters at runtime, forcing it to dump a raw netlog.json trace of every internal interaction the app was performing when I tapped on a song's lyrics.
From capturing those requests, I discovered a two-step API journey:

Step 1: Get the Lyrics browseId The first request hits the /next endpoint. I pass the currently playing videoId and use ANDROID_MUSIC as the clientName in the context payload.

POST https://music.youtube.com/youtubei/v1/next?prettyPrint=false
The JSON response returns a tree of layout elements. I parse through the JSON tree looking for the tabRenderer with the title "Lyrics". If lyrics exist for the track, that tab object will contain a specific browseEndpoint.browseId string.

Step 2: Fetch the actual lyrics data Once I have that specific browseId, I execute a second request to the /browse endpoint.

POST https://music.youtube.com/youtubei/v1/browse?prettyPrint=false
This payload is almost identical but uses the discovered browseId. The API responds with either a timedLyricsData array containing objects with lyricLine (text string) and cueRange.startTimeMilliseconds (for perfect syncing), or a standard paragraph block inside a node for non-synced tracks.

Example response from API:
image


harsh251004:
can you make it so that it uses the yt music synced lyrics provider just like the mobile app?

ShorMeneses:
> can you make it so that it uses the yt music synced lyrics provider just like the mobile app?

Done! Sorry it took some time, I was quite busy. Removed all LRCLib and Musixmatch provider, I had to dig a bit to understand what the mobile app was doing

harsh251004:
nice so you were able to do it?
can you send me the build?

ShorMeneses:
> nice so you were able to do it? can you send me the build?

It would be better if you built it by youself. (It's never safe to use binaries from strangers :laughing: )

However where it is:
[Build output](https://temp.sh/NLJFH/YouTube_Music_Desktop_App-win32-x64-2.0.11.zip)

nactrimo:
Tested this by rebasing onto current \`development\` (matches shipped v2.0.12) — it still merges and builds cleanly.

The synced lyrics fetch itself works great (verified via logs, correct timestamped data returned from the YT Music endpoint), but the UI never appeared for me: YouTube Music's web DOM has since dropped `#right-content` (now `#side-panel`) and no longer sets a `selected` attribute on `ytmusic-tab-renderer`, so `isLyricsTabSelected()` always returned `false`.

Found a reliable fix using `ytmusic-tab-renderer[page-type="MUSIC_PAGE_TYPE_TRACK_LYRICS"]`, which is set whenever the Lyrics tab is active regardless of UI language (also confirmed the old English-text tab matching wouldn't work on non-English locales either, e.g. French shows "Paroles" not "Lyrics"). Opened it against this branch here: ShorMeneses#1

With that fix, synced lyrics display and highlight correctly for regular audio tracks. One thing worth flagging: on official music-video ("clip") uploads, the synced timestamps sometimes don't match the video's edit (different intro/outro length than the canonical audio track), causing the line highlight to jump — but that matches the official YouTube Music mobile app's behavior on the same tracks, so it looks like an upstream data limitation rather than a bug in this PR.

Given the demand tracked in #809 / #1788 and that this uses YTM's own first-party lyrics data (no third-party ToS concerns), it'd be great to see this land. Happy to help however's useful.

Issue #114, opened 2020-03-08
Request AdBlock Support


cebl-hg:
Is it possible to add AdBlock Support within the app (maybe even as AddOn)?

phu54321:
Little skeptical that such movement may violate with Youtube's policy. I'm okay with another fork implementing that feature, but official repo recommending AdBlock doesn't seem great.

You should pay for the music you hear.

Margiris:
It was asked and rejected some time ago in #12 . I was not happy about that and since I didn't find any better YT Music apps than this one, I decided to add an adblocker myself:
https://github.com/margiris/ytmdesktop
Currently it's mostly run using npm, installations are only possible with a workaround in Arch. But if this fork gains enough attention I might release compiled binaries.

rickinbsb:
There is high chance that this would receive a DMCA if implemented, so this needs some discussion before implementing it upstream

debmaj2:
Hi! As you might have already known, Google is against Ad-Blocking mechanisms and we don't want this project to be ceased because of a notice from Google. If you wish to do Ad-blocking, you can do so by using Adguard Desktop Premium's Application Ad Blocking method or using Pi Hole. (Get a cheap deal on StackSocial for Adguard Premium: https://stacksocial.com/?rid=8646247)

Block81838:
its illegal

TotalChris:
You should try YouTube Music premium. You can get it bundled with YouTube Premium for $12 and it removes all YouTube ads for you. Another thing most people don't talk about is that YouTube will pay creators more directly if you watch with premium. If your favorite YouTuber has a demonetization problem, that might help them. That and it's only $2 more than Spotify.

Block81838:
> You should try YouTube Music premium. You can get it bundled with YouTube Premium for $12 and it removes all YouTube ads for you. Another thing most people don't talk about is that YouTube will pay creators more directly if you watch with premium. If your favorite YouTuber has a demonetization problem, that might help them. That and it's only $2 more than Spotify. 

It's just the same price as Spotify lol, 9.99usd/mo

flleeppyy:
Then how the hell has VancedApp managed to get away with having an option to remove ads in their app?

debmaj2:
> Then how the hell has VancedApp managed to get away with having an option to remove ads in their app?

Ask yourself the same question, how did Google ban several accounts using NewPipe.

[...]

PR #1668, opened 2026-02-18
feat: integrate ad blocker

rixabhh:
Implemented ad blocker using @ghostery/adblocker-electron.


Alipoodle:
https://github.com/ytmdesktop/ytmdesktop/issues/114#issuecomment-1732543750

While the work done is kindly appreciated. We have a clear stance on how we want to handle ad blocking within this project.

TL;DR. We don't want to add it part of the *core* code or promote existence with it. The idea of this project is to provide a product that Google/YouTube might make themselves.

PR #1689, opened 2026-03-07
ads the ad-block functionality

WyrdWonton:
if you're building yourself for whatever reason use "yarn make" in terminal


Alipoodle:
We have no plans of adding a native Adblocker to YouTube music Desktop app.

You can see more details per: #114

PR #1715, opened 2026-03-30
Add Share button with Telegram sharing to player bar

### LINKED ISSUE #452: [FEATURE] Telegram share button
- [ x ] I understand that **YTMDesktop have NO affiliation with Google or YouTube**.
- [ x ] I verified that there is no open issue for the same subject.
Is your feature request related to a problem? Please describe.

**Describe the solution you'd like**

A simple telegram icon in the "Share" Popup.
In detail there may be an API to directly access the Telegram-Application, so the desired chat to share with can be chosen.

jippylong12:
## Summary
- Adds a Share dropdown button to the player bar right controls (next to sleep timer)
- Telegram is the first share target -- clicking it opens the Telegram share page with the current song URL and title pre-filled
- Uses the existing YTM popup menu system (same pattern as sleep timer) so adding future share platforms is just adding a new menu item + handler case
- Allows `t.me` and `telegram.org` in the external URL allowlist

## Test plan
- [ ] Play a song, verify Share icon appears in the right controls area of the player bar
- [ ] Click Share button, verify dropdown menu appears with "Telegram" item
- [ ] Click Share button again while menu is open, verify it closes
- [ ] Click "Telegram", verify browser opens `t.me/share/url` with correct song URL and title
- [ ] Verify no errors when clicking Share with no song playing

Screenshot 2026-03-31 at 12 13 00

Screenshot 2026-03-31 at 12 12 20

Narrow Viewports

Screenshot 2026-03-31 at 12 13 39

Closes #452

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -909,7 +909,7 @@ function openExternalFromYtmView(urlString: string) {
   const domainSplit = url.hostname.split(".");
   domainSplit.reverse();
   const domain = `${domainSplit[1]}.${domainSplit[0]}`;
-  if (domain === "google.com" || domain === "youtube.com") {
+  if (domain === "google.com" || domain === "youtube.com" || domain === "t.me" || domain === "telegram.org") {
     shell.openExternal(urlString);
   }
 }
--- a/src/renderer/ytmview/preload.ts
+++ b/src/renderer/ytmview/preload.ts
@@ -14,6 +14,7 @@ import { StoreSchema } from "~shared/store/schema";
 
 import playerBarControlsScript from "./scripts/playerbarcontrols.script?raw";
 import hookPlayerApiEventsScript from "./scripts/hookplayerapievents.script?raw";
+import injectSharePanelScript from "./scripts/injectsharepanel.script?raw";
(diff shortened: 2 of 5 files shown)


harsh251004:
the share button should just open the default share menu instead of showing telegram.

jippylong12:
Thank you for the suggestion. Visually this is the cleaner path. I went with the custom button first because it was simpler and didn't require touching YTM's internal DOM, but the native dialog approach does work. It took a fair bit of iteration to get the layout matching the existing targets correctly.

A couple of things worth flagging about the implementation:
- Detection uses a MutationObserver plus a short periodic poll to catch when the share dialog renders
- The Telegram target clones an existing `yt-share-target-renderer` so it inherits YTM's flex layout, then waits a couple of frames for Polymer's data-binding to wipe the clone before swapping in our own inner content
- Styling is copied from YTM's computed values (padding, margin, display mode) rather than hand-rolled, so it matches visually

Seems to work in my testing. It does rely on a few YTM internal selectors so there's some inherent coupling.


Screenshot 2026-04-13 at 23 23 36

